PR #38, opened 2016-11-20
Suggestion to change the result object message

salameer:
from what I understand we provide an executive summary at the top and
now we're digging into the details, so maybe this message might be a bit
more precise


salameer:
I guess that now officially makes me a MAJOR contributor to this Repo

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

Issue #44, opened 2016-11-20
Create Centralized documentation for all of our Swagger validation tools


salameer:
The idea of this documentation is to point the Swagger author to all the swagger validation tools that he can use while authoring his swaggers, and provide:

- Clear and simple instructions to install the tools
- Clearly and simple description of how to execute the tool (with complete example commands if applicable)
- Clearly and simple examples of what type of issues would each tool catch. 
- At which step in the process of creating the swagger should every tool be used.

We can document this in the Wiki of this tool or create a documentation folder. This will also help us in our future integration with VS Code.

salameer:
@kirthik you can use this task to track the documentation efforts of our tools. the documentation can definitely be referenced from the Swagger checklist and eventually shared with the customers.

veronicagg:
@salameer is this issue still valid?

PR #59, opened 2016-12-16
Updating readme file

veronicagg:
Starting to address #44.


veronicagg:
@salameer please take a look and see if it addresses what you wanted as initial documentation. @amarzavery, it'd be great if you could review it too.

salameer:
Hi Veronica, I think it's really good start but I would suggest putting the introduction info that you added on top of the installation instructions :)

PR #70, opened 2017-03-04
add support for processing swaggers in YAML format

amarzavery:
@salameer - This should take care of SF's concerns.


salameer:
NICE!!!

PR #92, opened 2017-04-11
fix for headers with Content-Type

amarzavery:


lib/specValidator.js
```diff
@@ -555,7 +555,16 @@ class SpecValidator {
       }
     }
 
-    if (!(options.headers['content-type'] || options.headers['Content-Type'])) {
```
amarzavery: With the previous change, the tool throws an error `setting 'Content-Type of undefined` when headers is not defined (provided in the example) and the tool tries to set `options.headers['Content-Type'] = utils.getJsonContentType(operation.consumes);`.

Now have made the logic clear. If headers is provided then we check for lowercase content-type. If that is present then we delete that and set it to the expected header key. if the expected header does not have the value then we set it.

and if headers was not defined at all then we create an empty headers object inside options and then set the content-type.

veronicagg:
@amarzavery it'd be good to start adding test cases to this tool, this would be one case I'd recommend covering, so we make sure it doesn't regress. There may be also a need to add a test case on the sway side. thanks!

PR #99, opened 2017-04-16
code restructuring, spec changes, more tests, added liveRequestResponseValidation method

amarzavery:
- we still need more changes to the swagger spec. Will have a chat on Monday.


vishrutshah:
As discussed offline, Things looks good.

Issue #98, opened 2017-04-15
Work to be done on Monday


amarzavery:
- [x] Integrate operation.validateRequest()
- [x] Integrate operation.validateResponse()
- [x] LiveValidator buildCache() and lookUP() methods should throw LiveValidationError objects.
- [x] Integrate oav into express and run it on localhost:1337

amarzavery:
First two are done. Minor tweaks may be required after some discussion on Monday.

vishrutshah:
All done. closing this task

PR #101, opened 2017-04-17
Remodel getPotentialOperations operation to return reasons

vishrutshah:
Part of #98


vishrutshah:
@amarzavery I broke the build :)

PR #123, opened 2017-05-24
Prepare to release oav version 0.4.2

vishrutshah:
Fixes: https://github.com/Azure/oav/issues/122
+ 
Preparing for the new release


README.md
```diff
@@ -54,9 +54,19 @@ npm install -g oav
 ```
 bash-3.2$ oav -h
 Commands:
-  validate-example <spec-path>  Performs validation of x-ms-examples and
-                                examples present in the spec.
-  validate-spec <spec-path>     Performs semantic validation of the spec.
+  extract-xmsexamples <spec-path>           Extracts the x-ms-examples for a
```
veronicagg: I wonder if we should re-order to show the validate commands that we expect people to use most, first.

README.md
```diff
@@ -54,9 +54,19 @@ npm install -g oav
...
+  <recordings>                              given swagger from the .NET session
+                                            recordings and saves them in a file.
+  generate-wireformat <spec-path>           Transforms the x-ms-examples for a
+                                            given operation into raw
+                                            request/response format and saves
+                                            them in a markdown file.
+  resolve-spec <spec-path>                  Resolves the swagger spec based on
```
veronicagg: the explanation for this one may need more details, I'm not sure I get what it does or why I would use it from the text there. Do we need more documentation on some of these?

vishrutshah: How about `Resolves the swagger spec based on the selected options like allOfs, relativePaths, examples etc.` ?

README.md
vishrutshah: Currently by default it does sorting alphabetically. Quickly looking at [commandDir](https://github.com/yargs/yargs#commanddirdirectory-opts) I do not see the straight forward option yet. May be i'll just leave it for now.

vishrutshah:
Thanks for feedback and review!

PR #127, opened 2017-05-31
For 'extract-xmsexamples', allowing windows-style file-paths for '<sp…

manaas-microsoft:
…ec-path>' and '<recordings>'.

### DIFF
--- a/lib/xMsExampleExtractor.js
+++ b/lib/xMsExampleExtractor.js
@@ -69,7 +69,7 @@ class xMsExampleExtractor {
 
     let outputExamples = self.options.output + "/examples/";
     let relativeExamplesPath = "../examples/";
-    let specName = self.specPath.split("/");
+    let specName = self.specPath.split(/[/\\]/);
     let outputSwagger = self.options.output + "/swagger/" + specName[specName.length - 1].split(".")[0] + ".json"
 
     var swaggerObject = require(self.specPath);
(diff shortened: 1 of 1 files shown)


manaas-microsoft:
@veronicagg Hi, Was trying to generate x-ms-examples using this script. Since, I was using it on windows, hit issues related to path delimiters being "\\" instead of "/". Have made that change, and added here. Please do let me know if this looks fine.

lib/xMsExampleExtractor.js
```diff
@@ -149,7 +149,7 @@ class xMsExampleExtractor {
                   let infoFromOperation = paths[path][requestMethodFromRecording.toLowerCase()];
                   if (typeof infoFromOperation != 'undefined') {
                     //need to consider each method in operation
-                    let fileName = recordingFileName.split('/');
+                    let fileName = recordingFileName.split(/[/\\]/);
```
vishrutshah: Should you use `pathlib.sep` instead?

lib/xMsExampleExtractor.js
```diff
@@ -69,7 +69,7 @@ class xMsExampleExtractor {
 
     let outputExamples = self.options.output + "/examples/";
     let relativeExamplesPath = "../examples/";
-    let specName = self.specPath.split("/");
+    let specName = self.specPath.split(/[/\\]/);
```
vishrutshah: same question here

lib/xMsExampleExtractor.js
manaas-microsoft: Sure. Will check that. But I wanted to allow both kind of inputs "windows-style" and "Linux-style" to be parsed successfully. Will check if that happens using above option.

amarzavery:
closing this PR due to age. Please feel free to open if required.

PR #131, opened 2017-06-26
support yaml format

amarzavery:


amarzavery:
@bradygaster - Please take a look at this PR.
Way to generate wireformat as a yaml doc
```bash
oav generate-wireformat <spec-path> -y
```

lib/commands/generate-wireformat.js
```diff
@@ -24,6 +24,12 @@ exports.builder = {
     'need to be transformed. If operationIds are not provided then the entire spec will be processed. ' +
     'Example: "StorageAccounts_Create, StorageAccounts_List, Usages_List".',
     string: true
+  },
+  y: {
+    alias: 'inYaml',
+    describe: 'A boolean flag when provided will indicate the tool to ' +
+    'generate wireformat in a yaml doc. Default is a markdown doc.',
+    boolean: true
```
vishrutshah: minor: we can put ` default: true` here so that we do not have to initialize it at https://github.com/Azure/oav/pull/131/files#diff-91f3c7f79a6e7627c03b9c350ee8a9b2R13

PR #137, opened 2017-07-11
Live validator changes to support re-orged repo: azure-rest-api-specs

vishrutshah:


lib/validators/liveValidator.js
```diff
@@ -98,7 +98,8 @@ class LiveValidator {
       swaggerPaths = self.options.swaggerPaths;
       log.debug(`Using user provided swagger paths. Total paths: ${swaggerPaths.length}`);
     } else {
-      swaggerPaths = glob.sync(path.join(self.options.directory, '/**/swagger/*.json'));
+      let allJsons = path.join(self.options.directory, '/specification/**/*.json');
```
amarzavery: if this works -> `'/specification/**/swagger/**/*.json'` then you don't need to do filtering twice

vishrutshah: @amarzavery Now swagger files don't belong to folder name `swagger`. So that should not work for us :)

Issue #139, opened 2017-07-12
Make oav more "pluggable" in autorest for consumption in VS Code extension


veronicagg:
- Make oav more of a plugin to autorest
- Remove code in autorest that involves logic and interpretation of oav
This is to enable autorest to call oav for model validation and semantic validation, which is currently broken, and enable exposing these messages in VS Code extension.

PR #143, opened 2017-07-25
Preparing oav to plug-in to autorest 

veronicagg:
Preparing oav to plugin to autorest and adding id to error codes for consistency and documentation.
Creating pluginHost.js to enable plugging in to AutoRest model. 
Created IDs for error codes for consistency reporting errors with "linter" tool; and referencing in documentation to help trouble shoot the issues. I'll send a PR later on with corresponding documentation.

This PR includes just example validation integration, I'll be sending another one or adding to this one for semantic validation. 

Addressing issue https://github.com/Azure/oav/issues/139


lib/autorestPlugin/stable-object.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
```
olydis: 99% sure you don't need this file

lib/autorestPlugin/yaml.js
```diff
@@ -0,0 +1,177 @@
+"use strict";
```
olydis: 99% sure you don't need this file

lib/autorestPlugin/pluginHost.js
```diff
@@ -0,0 +1,204 @@
...
+exports = module.exports;
+
+class OpenApiValidationExample {
+  Process(sessionId, initiator) {
+    return initiator.ListInputs(sessionId).then(function (swaggerFileNames) {
+      const promises = [];
+      for (const swaggerFileName of swaggerFileNames) {
+        promises.push(
+          initiator.ReadFile(sessionId, swaggerFileName).then(function (swaggerFile) {
+            const swagger = yaml.Parse(swaggerFile);
```
olydis: just use `js-yaml`'s `safeLoad` here. The homebrew parser does way more stuff (source map generation helpers, stable order when round tripping objects like `{ b: 3, a: 2 }` vs `{ a: 2, b: 3 }`) but all that is super irrelevant here 😉

lib/autorestPlugin/pluginHost.js
```diff
@@ -0,0 +1,204 @@
+/*---------------------------------------------------------------------------------------------
```
olydis: for code that is all about logic and not about protocol, you can use this: https://github.com/olydis/autorest-extension-base (see README, it's just 3 lines of overhead)

vishrutshah:
@veronicagg Awesome!! We'll now be having validation through autorest 👍 👍  As i am not super duper familiar with the plugin thingy would you please setup sometime to go over this changes to help me review? On the other hand, some changes would require validating the live validation works as well (mostly yes). Thanks!

lib/autorestPlugin/pluginHost.js
```diff
@@ -0,0 +1,204 @@
...
+  return result;
+}
+
+class PluginHost {
+  constructor(initiator) {
+    this.initiator = initiator;
+  }
+
+  GetPluginNames() {
+    return [/*OpenApiValidationSemantic.Name, */ OpenApiValidationExample.Name];
```
dsgouda: nit: why is this commented?

veronicagg: Oh, it's commented out, because I haven't hooked up Semantic validation yet, it's coming...

lib/autorestPlugin/pluginHost.js
veronicagg: cool, I'll try that, so I can remove the 2 files above.

lib/autorestPlugin/pluginHost.js
veronicagg: sure, will incorporate it, thanks!

veronicagg:
@vishrutshah thanks, yes, the live validator tests passed, but it'd be great to check on this further with you.

vishrutshah:
LiveValidator need to be updated https://github.com/Azure/oav/blob/master/lib/validators/liveValidator.js#L225, we need to update `Constants.ErrorCodes`

lib/models/liveValidationError.js
```diff
@@ -21,7 +21,7 @@
  */
 class LiveValidationError {
   constructor(code, message) {
-    this.code = code;
+    this.code = code.name;
```
vishrutshah: ooppss I wouldn't change this. As this is auto generated from Autorest from swagger.

vishrutshah: We may need to change the callers to pass on the code as as a string

veronicagg: thanks Vishrut, makes sense, I went with the simplest approach for now and updated the callers to pass the name.

vishrutshah: 👍

vishrutshah:
@veronicagg Change LGTM! 

One thing i noticed (sorry may be I missed previously) is that we are checking-in `package-lock.json` file. I think which is the right thing to do but that file lists version of oav: 0.4.3 which is pretty old. Can you please make sure to remove the file and do `npm install` so that we have upto date lock file. Thanks!

veronicagg:
@vishrutshah @olydis @dsgouda thanks all for the review!!!

Issue #165, opened 2017-09-13
Validate examples reporting error when there are "/" values for path parameters


veronicagg:
Here's the example where this issue occurred: 
-	Swagger def of param: https://github.com/Azure/azure-rest-api-specs/pull/1648/files#diff-eb84d1cad45c568273be52d4cec4c047R33 
-	Example def with param: https://github.com/Azure/azure-rest-api-specs/pull/1648/files#diff-196fec1ea9372d75cba90e958b3ee8d1R3 
-	CI complaining about missing param: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/273126090#L859

https://github.com/Azure/azure-rest-api-specs/pull/1648

veronicagg:
Error:

{ code: 'REQUEST_VALIDATION_ERROR',     id: 'OAV109',     message: 'Found errors in validating the request   for x-ms-example "EventSubscriptions_CreateForSubscription" in   operation "EventSubscriptions_Create".',     innerErrors:      [ { code: 'INVALID_REQUEST_PARAMETER',          errors:           [ { code:   'REQUIRED',                 message: 'Value is   required but was not provided',                 path:                  [ 'paths',                    '/{scope}/providers/Microsoft.EventGrid/eventSubscriptions/{eventSubscriptionName}',                    'put',                    'parameters',                    '0' ] } ],          in: 'path',          message: 'Invalid parameter (scope):   Value is required but was not provided',          name: 'scope',          path: { '$ref':   '$["innerErrors"][0]["errors"][0]["path"]' } },        { code:   'INVALID_REQUEST_PARAMETER',          errors:           [ { code:   'REQUIRED',                 message: 'Value is required but was not provided',                 path:                  [ 'paths',                    '/{scope}/providers/Microsoft.EventGrid/eventSubscriptions/{eventSubscriptionName}',                    'put',                    'parameters',                    '1' ] } ],          in: 'path',          message: 'Invalid parameter   (eventSubscriptionName): Value is required but was not provided',          name:   'eventSubscriptionName',          path: { '$ref':   '$["innerErrors"][1]["errors"][0]["path"]' } }   ] }
--

amarzavery:
fixed.

Issue #140, opened 2017-07-14
oav reports error for example validation when spec uses x-ms-paths 


veronicagg:
Error:
error: Found errors in validating the request for x-ms-example "Sets the owner of a file or directory" in operation "FileSystem_SetPermission".:

{ code: 'REQUEST_VALIDATION_ERROR',
  message: 'Found errors in validating the request for x-ms-example "Sets the owner of a file or directory" in operation "FileSystem_SetPermission".',
  innerErrors:
   [ { code: 'INVALID_REQUEST_PARAMETER',
       errors:
        [ { code: 'REQUIRED',
            message: 'Value is required but was not provided',
            path:
             [ 'paths',
               '/webhdfs/v1/{path}?op=SETPERMISSION',
               'put',
               'parameters',
               '0' ] } ],
       in: 'path',
       message: 'Invalid parameter (path): Value is required but was not provided',
       name: 'path',
       path: { '$ref': '$["innerErrors"][0]["errors"][0]["path"]' } } ] }

Based on investigation '?' is confusing the model validation at:
https://github.com/amarzavery/sway/blob/validation/lib/types/parameter.js#L164

Instead of reaching that point with '?' which is basically being included because of "x-ms-paths" and its "autorest-isms", we can replace '?' with '/' for the purposes of example validation when unifying x-ms-paths. This is also a local workaround to validate examples, until this fix is done.

veronicagg:
we should check that this doesn't cause a problem for the live-validation switch

amarzavery:
fixed.

PR #166, opened 2017-09-13
fixes for #140 and #165

amarzavery:
fixes for #140 and #165

### DIFF
--- a/lib/validators/specValidator.js
+++ b/lib/validators/specValidator.js
@@ -528,14 +528,6 @@ class SpecValidator {
         resultScenarios[scenario] = {};
         resultScenarios[scenario].requestValidation = self.validateRequest(operation, xmsExample.parameters);
         resultScenarios[scenario].responseValidation = self.validateXmsExampleResponses(operation, xmsExample.responses);
-        // TODO: Usefult for creating live request and response from x-ms-examples
-        // let fileName = path.join('/Users/amarz/sdk/openapi-validation-tools/test/swaggers/arm-storage/2016-01-01/live',
-        //   `${operation.operationId}.json`);
-        // let data = {
-        //   liveRequest: self.sampleRequest,
-        //   liveResponse: self.sampleResponse
-        // }
-        // fs.writeFileSync(fileName, JSON.stringify(data, null, 2), { encoding: 'utf8' });
       }
     } else {
       let msg = `x-ms-example not found in ${operation.operationId}.`;
@@ -590,7 +582,8 @@ class SpecValidator {
     options.pathTemplate = operation.pathObject.path;
     for (let i = 0; i < parameters.length; i++) {
       let parameter = parameters[i];
-      if (!exampleParameterValues[parameter.name]) {
+      let parameterValue = exampleParameterValues[parameter.name];
+      if (!parameterValue) {
         if (parameter.required) {
           let msg = `In operation "${operation.operationId}", parameter ${parameter.name} is required in the ` +
             `swagger spec but is not present in the provided example parameter values.`;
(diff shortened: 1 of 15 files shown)


amarzavery:
@vishrutshah and @veronicagg please review.

test/liveValidatorTests.js
```diff
@@ -117,7 +117,7 @@ describe('Live Validator', function () {
       let validator = new LiveValidator(options);
       validator.initialize().then(function () {
         assert.equal(true, expectedProvider in validator.cache);
-        assert.equal(5, Object.keys(validator.cache).length);
+        assert.equal(6, Object.keys(validator.cache).length);
```
vishrutshah: QQ: Is this because of the test.json or the fix ?

amarzavery: yes. I could have added this in a separate folder. But then thought of keeping all the swaggers together. In that way we know that live validator will work well for the newly added edge case scenario tests as well.

lib/validators/specValidator.js
```diff
@@ -603,22 +596,35 @@ class SpecValidator {
...
       let location = parameter.in;
       if (location === 'path' || location === 'query') {
+        if (location === 'path' && parameterValue && typeof parameterValue.valueOf() === "string") {
+          if (options.pathTemplate.charAt(options.pathTemplate.indexOf(`${parameter.name}`) - 2) === '/' && parameterValue.startsWith('/')) {
+            let msg = `In operation "${operation.operationId}", example for parameter "${parameter.name}": "${parameterValue}" starts with a forward slash ` +
+              `and the path template: "${options.pathTemplate}" contains a forward slash before the parameter starts. This will cause double forward slashes ` +
+              ` in the request url. Thus making it incorrect. Please rectify the example.`;
+            let e = self.constructErrorObject(ErrorCodes.DoubleForwardSlashesInUrl, msg);
+            result.validationResult.errors.push(e);
+            foundIssues = true;
```
amarzavery: I am still iffy about this. Not sure if throwing an error for this scenario is necessary.

vishrutshah: I think throwing would give more visibility than swallowing...

lib/validators/specValidator.js
```diff
@@ -603,22 +596,35 @@ class SpecValidator {
       }
       let location = parameter.in;
       if (location === 'path' || location === 'query') {
+        if (location === 'path' && parameterValue && typeof parameterValue.valueOf() === "string") {
+          if (options.pathTemplate.charAt(options.pathTemplate.indexOf(`${parameter.name}`) - 2) === '/' && parameterValue.startsWith('/')) {
```
veronicagg: what's "-2" looking for here? it'd be great to add a small comment potentially with an example on why we're adding this condition. There's a message you're constructing right after, but a comment wouldn't hurt for readability.

lib/validators/specValidator.js
```diff
@@ -603,22 +596,35 @@ class SpecValidator {
...
+            let msg = `In operation "${operation.operationId}", example for parameter "${parameter.name}": "${parameterValue}" starts with a forward slash ` +
+              `and the path template: "${options.pathTemplate}" contains a forward slash before the parameter starts. This will cause double forward slashes ` +
+              ` in the request url. Thus making it incorrect. Please rectify the example.`;
+            let e = self.constructErrorObject(ErrorCodes.DoubleForwardSlashesInUrl, msg);
+            result.validationResult.errors.push(e);
+            foundIssues = true;
+            break;
+          }
+          //replacing forward slashes with empty string because this messes up Sways regex validation of path segment.
+          parameterValue = parameterValue.replace(/\//ig, '');
```
veronicagg: I found https://github.com/OAI/OpenAPI-Specification/issues/892. It appears that having path parameters with slashes may not be allowed in swagger??

lib/validators/specValidator.js
veronicagg: If we throw we stop any other validation, right? If we log an errors, we report it as any other? If so, I'd go with error.

test/swaggers/specification/scenarios/resource-manager/Microsoft.Test/2016-01-01/examples/storageAccountPathParameterForwardSlash.json
```diff
@@ -0,0 +1,12 @@
+{
+  "parameters":{
+    "api-version": "2016-01-01",
+    "subscriptionId": "foo/foo1/foo2",
```
veronicagg: we could make it a little bit more real, if we had another parameter other than subscriptionId, since subscriptionID does not follow this pattern.

test/swaggers/specification/scenarios/resource-manager/Microsoft.Test/2016-01-01/test.json
```diff
@@ -0,0 +1,224 @@
...
+  "paths": {
+    "/subscriptions/{subscriptionId}/providers/Microsoft.Test/checkNameAvailability": {
+      "post": {
+        "tags": [
+          "StorageAccounts"
+        ],
+        "operationId": "StorageAccounts_CheckNameAvailability",
+        "description": "Checks that the storage account name is valid and is not already in use.",
+        "x-ms-examples": {
+          "storageAccountCheckNameAvailability": {"$ref": "./examples/storageAccountCheckNameAvailability.json"}
```
veronicagg: you could probably have multiple examples for 2 operations (one for the forward slashes, one of the "?"), instead of multiple operations with one example each, it may simplify test.json.

lib/validators/specValidator.js
amarzavery: I think my sentence in the comment confused you guys. Sorry about that
> I am still iffy about this. Not sure if throwing an error for this scenario is necessary.

Yes we are just logging the error like we do for other scenarios. We are not throwing and halting further validation. 
What I actually meant to ask is 
> "Does it actually make sense to catch these kind of issues? I mean in the real world customers could provide forward slashes in their parameter values when they use an sdk or the portal or REST API directly. Shouldn't this be handled by the service. (they should ignore duplicate forward slashes or put a regex validation constraint in swagger spec that the value should not start with forward slash". Making this as a brute force validation in the model validator is too strict.

What do you think?

lib/validators/specValidator.js
amarzavery: Yes they are technically not allowed. But in real world services have such urls and we need to support them.
Think about apis like dropbox or file storage apis where they have a file path or container path as a path parameter. They wont be able to use swagger.

test/swaggers/specification/scenarios/resource-manager/Microsoft.Test/2016-01-01/test.json
amarzavery: I wanted to provide a scenario that has both the issues and verify whether our tool correctly catches it.

amarzavery:
https://github.com/amarzavery/sway/pull/9

lib/validators/specValidator.js
veronicagg: yeah, I thought about that while reviewing too, I think we could just swallow honestly for now and fix it "silently", I think having a warning instead of an error could be a good thing, in case we decide to use this value for something else later and the impact is different - we would at least know it's flagged.

PR #169, opened 2017-09-25
Provide the filename in error message that has an incorrect json.

amarzavery:


amarzavery:
@vishrutshah please review

Issue #148, opened 2017-08-07
Make Model Validation Output Similar to Semantic and Linter Output


shawns1:
Model Validation format changes to improve data collection:
There are some validation output changes that would make data collection easier.  The primary issues when collecting data are that there isn’t a standard format for each error returned and that the json file is broken down by node names for the errors rather than something more generic like arrays like the Semantic validator output.  Examples of both outputs are below.
Currently linter and semantic validation:  It is an array of errors per spec
 
{
  "D:/workspace/azure-rest-api-specs-Validation/specification/advisor/resource-manager/Microsoft.Advisor/2016-07-12-preview/advisor.json": [
    {
      "type": "Warning",
      "code": "OperationIdNounVerb",
      "message": "Per the Noun_Verb convention for Operation Ids, the noun 'Recommendations' should not appear after the underscore. Note: If you have already shipped an SDK on top of this spec, fixing this warning may introduce a breaking change.",
      "id": "R1001",
      "validationCategory": "SDKViolation",
      "providerNamespace": "Microsoft.Advisor",
      "resourceType": "generateRecommendations",
      "sources": [
        "file:///D:/workspace/azure-rest-api-specs-Validation/specification/advisor/resource-manager/Microsoft.Advisor/2016-07-12-preview/advisor.json:51:4 ($.paths[\"/subscriptions/{subscriptionId}/providers/Microsoft.Advisor/generateRecommendations/{operationId}\"].get.operationId)"
      ]
    },
    {
      "type": "Error",
      "code": "BodyTopLevelProperties",
      "message": "Top level properties should be one of name, type, id, location, properties, tags, plan, sku, etag, managedBy, identity. Model definition 'SuppressionContract' has extra properties ['suppressionId,ttl'].",
      "id": "R3006",
      "validationCategory": "RPCViolation",
      "providerNamespace": null,
      "resourceType": null,
      "sources": [
        "file:///D:/workspace/azure-rest-api-specs-Validation/specification/advisor/resource-manager/Microsoft.Advisor/2016-07-12-preview/advisor.json:583:3 ($.definitions.SuppressionContract.properties)"
      ]
    }
  ],
  "D:/workspace/azure-rest-api-specs-Validation/specification/advisor/resource-manager/Microsoft.Advisor/2017-03-31/advisor.json": [
    {…
 
 
 
Modeler is based on node names which are mapped to object names when de-serializing.  
 
 
{
[...]

PR #170, opened 2017-10-02
Updating model validator plugin to autorest

veronicagg:
Based on latest guidance from autorest, updating plugin configuration. 
Also, addressing formatting out output based on https://github.com/Azure/oav/issues/148

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,5 +82,17 @@ Options:
 bash-3.2$
 ```
 
+### Autorest plugin configuration
+- Please don't edit this section unless you're re-configuring how oav plugs in to AutoRest
+AutoRest needs the below config to pick this up as a plug-in - see https://github.com/Azure/autorest/blob/master/docs/developer/architecture/AutoRest-extension.md
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -82,5 +82,17 @@ Options:
...
 
+### Autorest plugin configuration
+- Please don't edit this section unless you're re-configuring how oav plugs in to AutoRest
+AutoRest needs the below config to pick this up as a plug-in - see https://github.com/Azure/autorest/blob/master/docs/developer/architecture/AutoRest-extension.md
+
+``` yaml $(model-validator)
+pipeline:
+  swagger-document/model-validator:
+    input: swagger-document/identity
+    scope: azure-validator-composed
```
olydis: think you can get rid of that - even more so, I don't see a scope called `azure-validator-composed` in here, that was necessary for Deepak distinction between rules for single vs. merged Swagger - model validator doesn't have such distinction

olydis: @veronicagg any plans on moving forward with this?

Issue #142, opened 2017-07-19
Polymorphic operation validation fails - 


vishrutshah:
```
D:\Repos\oav>oav validate-example "C:\\Users\\vissha\\Downloads\\2017-03-01-preview\\swagger\\datafactory.json" -o Pipelines_ListByFactory
error: Found errors in validating the response with statusCode "200" for x-ms-example "Pipelines_ListByFactory" in operation "Pipelines_ListByFactory".:

{ code: 'RESPONSE_VALIDATION_ERROR',
  message: 'Found errors in validating the response with statusCode "200" for x-ms-example "Pipelines_ListByFactory" in operation "Pipelines_ListByFactory".',
  innerErrors:
   [ { code: 'INVALID_RESPONSE_BODY',
       errors:
        [ { code: 'OBJECT_ADDITIONAL_PROPERTIES',
            params: [ [ 'typeProperties' ] ],
            message: 'Additional properties not allowed: typeProperties',
            path:
             [ 'value',
               '0',
               'properties',
               'activities',
               '0',
               'typeProperties',
               'activities',
               '0' ],
            description: 'A pipeline activity.' } ],
       message: 'Invalid body: Additional properties not allowed: typeProperties',
       path: [] } ] }
```
Working `amarzavery/sway` commit: [3429105b265070d9fed0033528e02894a2540733](https://github.com/amarzavery/sway/tree/3429105b265070d9fed0033528e02894a2540733)

vishrutshah:
Child Activity types are not recognized for Container ..

This need to change the logic. We should traverse the `x-ms-examples` and then match it against the swagger schema to support multi-level polymorphic examples.

amarzavery:
This issue can be resolved by processing the model with discriminator property. If we replace all the references to the parent model with a oneOf array containing references to the parent model and all its child models and modify the discriminator property in the child models by making it a constant (enum with one value) with the value expected for that model type on the wire (model name or whatever is indicated by x-ms-discriminator-value). 
 

For example: There is a model named "Animal" with a discriminator as "animalType". Models like "Cat", "Dog", Tiger" are children (having "allof": [ { "$ref": "#/definitions/Animal" } ] on) of "Animal" in the swagger spec. 

 - We will have to replace all the locations in the swagger spec that have a reference to the
 parent model "Animal" ("$ref": "#/definitions/Animal") with a oneOf reference 
 ```
[...]

PR #171, opened 2017-10-17
Resolve the polymorphic models and validate examples

vishrutshah:
Part of https://github.com/Azure/oav/issues/142.

This issue can be resolved by processing the model with discriminator property. If 
- we replace all the references to that parent model with a oneOf array containing references to the parent model and all its child models 
- and modify the discriminator property in the child models by making it a constant (enum with one value) with the value expected for that model type on the wire (model name or whatever is indicated by x-ms-discriminator-value). (This can be done only after we resolve allof)*.
 

For example: There is a model named "Animal" with a discriminator as "animalType". Models like "Cat", "Dog", Tiger" are children (having "allof": [ { "$ref": "#/definitions/Animal" } ] on) of "Animal" in the swagger spec. 

 - We will have to replace all the locations in the swagger spec that have a reference to the
 parent model "Animal" ("$ref": "#/definitions/Animal") with a oneOf reference.
**NOTE:** We also need to make sure that we **_do not_** modify the allOf reference that a child model will have on it's parent. 
```json5
 "oneOf": [ 
  { "$ref": "#/definitions/Animal" },
  { "$ref": "#/definitions/Cat" },
  { "$ref": "#/definitions/Dog" },
  { "$ref": "#/definitions/Tiger" }
]
``` 
 - We will also add a constant value (name of that animal on the wire or the value provided by 
 "x-ms-discriminator-value") to the discrimiantor property "animalType" for each of the child models. 
 For example:  the Cat model's discriminator property will look like:
```json5
"Cat": { 
  "required": [ "animalType" ], 
  "properties": { 
    "animalType": { 
      "type": "string",
      "enum": [ "Cat" ]
     }, 
     . . . 
  }
}
```
 
**NOTE:** We have tested this on polymorphic collections that were circular as well and were able to validate such models correctly.


lib/validators/specResolver.js
```diff
@@ -268,7 +281,7 @@ class SpecResolver {
                 unresolvedDefinitions.push(() => {
                   if (result.definitions[defName].allOf) {
                     let matchFound = result.definitions[defName].allOf.some((item) => {
-                      return (item['$ref'] === `#${slicedLocalReferenceValue}` && !self.visitedEntities[`/definitions/${defName}`]);
+                      return (!self.visitedEntities[`/definitions/${defName}`]);
```
vishrutshah: why did we remove this? - Add comment like: Did not make sense of checking `slicedLocalReferenceValue` with the current `$ref` as they may or may not be same.

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+        for (let key in references) {
+          if (references[key].uri === modelReference && key.indexOf("allOf") === -1) locationsToBeUpdated.push(key);
+        }
+        for (let location of locationsToBeUpdated) {
+          utils.setObject(spec, `${location.slice(1)}`, { "oneOf": this.buildOneOfReferences(root) });
+        }
+      }
+    });
+    // delete definitions['Pipeline']['properties']['activities']['items']['$ref'];
+    // definitions['Pipeline']['properties']['activities']['items'].oneOf = [
```
vishrutshah: remove this

lib/util/utils.js
```diff
@@ -314,7 +315,7 @@ exports.mergeObjects = function mergeObjects(source, target) {
       }
       target[key] = exports.mergeArrays(source[key], target[key]);
     } else {
-      target[key] = source[key];
+      target[key] = lodash.cloneDeep(source[key]);
```
vishrutshah: clone as we do not want the references around as we may modify them independent of each other later.

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+ */
+  resolveDiscriminator() {
+    let self = this;
+    let spec = self.specInJson;
+    let definitions = spec.definitions;
+    let modelNames = Object.keys(definitions);
+    let subTreeMap = new Map();
+    let references = JsonRefs.findRefs(spec);
+
+    modelNames.map((modelName) => {
```
vishrutshah: add docs and tests

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+    for (let entry of root.children.entries()) {
+      if (entry[1]) {
+        result = result.concat(this.buildOneOfReferences(entry[1]));
+      }
+    }
+    return result;
+  }
 }
 
+class PolymorphicTree {
```
vishrutshah: :wow:

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+        let modelReference = `#/definitions/${modelName}`;
+        let root = subTreeMap.get(modelName), locationsToBeUpdated = [];
+        if (!root) {
+          root = this.createPolymorphicTree(modelName, definitions[modelName].discriminator, subTreeMap);
+        }
+        for (let key in references) {
+          if (references[key].uri === modelReference && key.indexOf("allOf") === -1) locationsToBeUpdated.push(key);
+        }
+        for (let location of locationsToBeUpdated) {
+          utils.setObject(spec, `${location.slice(1)}`, { "oneOf": this.buildOneOfReferences(root) });
```
amarzavery: This statement removes the description property along with the $ref that we actually want to remove. Hence, performing the following steps would be a better approach:
- doing a `utils.getObject()`
- deleting the `$ref` property
- adding the `oneOf` array of references comprising of the parent and the child
- doing a `utils.setObject()` with the updated object

```javascript
for (let location of locationsToBeUpdated) {
  const slicedLocation = location.slice(1);
  let obj = utils.getObject(spec, slicedLocation);
  if (obj) {
    if (obj['$ref']) delete obj['$ref'];
    obj.oneOf = this.buildOneOfReferences(root);
    utils.setObject(spec, slicedLocation, obj);
  }
}
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+
+    if (subTreeMap === null || subTreeMap === undefined || !(subTreeMap instanceof Map)) {
+      throw new Error('subTreeMap is a required property of type Map.')
+    }
+
+    let root = new PolymorphicTree(name);
+    let definitions = this.specInJson.definitions;
+
+    // Adding the model name or it's discriminator value as an enum constraint on property marked as discriminator
+    if (definitions[name] && definitions[name].properties && definitions[name].properties[discriminator]) {
```
amarzavery: // Adding the model name or it's discriminator value as an enum constraint **with one value (constant)** on property marked as discriminator

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+  buildOneOfReferences(root) {
+    let result = [{ "$ref": `#/definitions/${root.name}` }];
+    for (let entry of root.children.entries()) {
+      if (entry[1]) {
+        result = result.concat(this.buildOneOfReferences(entry[1]));
+      }
+    }
+    return result;
+  }
 }
```
amarzavery:
```javascript
/**
 * @class
 * Creates a tree by traversing the definitions where the parent model is the root and child model is one of it's children.
 */
```

lib/validators/specResolver.js
amarzavery:
```javascript
/**
 * @constructor
 * Initializes a new instance of the PolymorphicTree
 *
 * @param {string} name - The name of the parent model
 * @param {Map<string, PolymorphicTree>} [children] - A map of zero or more children representing the child models in the inheritance chain
 */

File: lib/validators/specResolver.js [hunk 9]
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+      throw new Error('name is a required property of type string and it cannot be an empty string.')
+    }
+
+    if (children !== null && children !== undefined && !children instanceof Map) {
+      throw new Error('children is an optional property of type Map<string, PolymorphicTree>.')
+    }
+    this.name = name;
+    this.children = children || new Map();
+  }
+
```
amarzavery:
```javascript
/**
 * Adds a child by name to the PolymorphicTree. This method will not add the child again if it is already present.
 *
 * @param {string} childName- The name of the child model
 * @returns {PolymorphicTree} child - The created child node.
 */
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+    let child;
+    if (!this.children.has(childName)) {
+      child = new PolymorphicTree(childName);
+      this.children.set(childName, child);
+    } else {
+      child = this.children.get(childName);
+    }
+    return child;
+  }
+
```
amarzavery:
```javascript
/**
 * Adds a childObject to the PolymorphicTree. This method will not add the child again if it is already present.
 *
 * @param {PolymorphicTree} childObj- A polymorphicTree representing the child model.
 * @returns {PolymorphicTree} childObj - The created child node.
 */
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+    //   },
+    //   {
+    //     "$ref": "#/definitions/DataLakeAnalyticsUSQLActivity"
+    //   }
+    // ];
+
+    console.log("Hello");
+    return Promise.resolve(self);
+  }
+
```
amarzavery:
```javascript
/**
 * Creates a PolymorphicTree for a given model in the inheritance chain
 *
 * @param {string} name- Name of the model for which the tree needs to be created.
 * @param {string} discriminator- Name of the property that is marked as the discriminator.
 * @param {Map<string, PolymorphicTree>} subTreeMap - A map that stores a reference to PolymorphicTree for a given model in the inheritance chain.
 * @returns {PolymorphicTree} root- A PolymorphicTree that represents the model in the inheritance chain.
 */
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+    let children = this.findChildren(name);
+    for (let childName of children) {
+      let childObj = this.createPolymorphicTree(childName, discriminator, subTreeMap);
+      root.addChildByObject(childObj);
+    }
+    //Adding the created sub tree in the subTreeMap for future use.
+    subTreeMap.set(root.name, root);
+    return root;
+  }
+
```
amarzavery:
```javascript
/**
 * Finds children of a given model in the inheritance chain.
 *
 * @param {string} name- Name of the model for which the children need to be found.
 * @returns {Set} result- A set of model names that are the children of the given model in the inheritance chain.
 */
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
...
+            log.debug(`reference found: ${reference} in definition: ${definitionName}`);
+            result.add(definitionName);
+          }
+        });
+      }
+    }
+    return result;
+  }
+
+  buildOneOfReferences(root) {
```
amarzavery:
```javascript
/**
 * Builds the oneOf array of references that comprise of the parent and its children.
 *
 * @param {PolymorphicTree} root- A PolymorphicTree that represents the model in the inheritance chain.
 * @returns {PolymorphicTree} result- An array of reference objects that comprise of the parent and its children.
 */
```

lib/validators/specResolver.js
```diff
@@ -488,6 +501,281 @@ class SpecResolver {
     }
     return Promise.resolve(self);
   }
+
+  /**
+ * Deletes all the references to allOf from all the model definitions in the swagger spec.
+ */
```
amarzavery:
```javascript
/**
 * Resolves the discriminator by replacing all the references to the parent model with a oneOf array containing
 * references to the parent model and all its child models. It also modifies the discriminator property in
 * the child models by making it a constant (enum with one value) with the value expected for that model 
 * type on the wire. 
 * For example: There is a model named "Animal" with a discriminator as "animalType". Models like "Cat", "Dog",
 * Shark" are children (having "allof": [ { "$ref": "#/definitions/Animal" } ] on) of "Animal" in the swagger spec. 
 *
 * - This method will replace all the locations in the swagger spec that have a reference to the
 * parent model "Animal" ("$ref": "#/definitions/Animal") except the allOf reference with a oneOf reference 
 * "oneOf": [ { "$ref": "#/definitions/Animal" }, { "$ref": "#/definitions/Cat" },  { "$ref": "#/definitions/Dog" }, { "$ref": "#/definitions/Tiger" }  ]
 *
 * - It will also add a constant value (name of that animal on the wire or the value provided by 
 * "x-ms-discriminator-value") to the discrimiantor property "animalType" for each of the child models. 
 * For example:  the Cat model's discriminator property will look like:
 * "Cat": { "required": [ "animalType" ], "properties": { "animalType": { "type": "string", "enum": [ "Cat" ] },  . . } }.
 */
```

vishrutshah: 👍

lib/validators/specResolver.js
vishrutshah: 👍

lib/validators/specResolver.js
vishrutshah: done

lib/validators/specResolver.js
vishrutshah: 👍

PR #176, opened 2017-10-23
less verbose messages and no newline in the end

olydis:
Messages are too verbose (contain JSON blobs) for regular runs and only provide information that is also in the `details` field (where it belongs :wink: ).

Now:
```
ERROR (RESPONSE_VALIDATION_ERROR/OAV108/ExampleModelViolation): Invalid body: Value failed JSON Schema validation.
Scenario: Create/Update data masking rule for default min
More info: https://github.com/Azure/oav#oav108-response_validation_error
  - file:///c:/work/azure-rest-api-specs/specification/sql/resource-manager/Microsoft.Sql/2014-04-01/dataMasking.json:157:20 ($.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/databases/{databaseName}/dataMaskingPolicies/{dataMaskingPolicyName}/rules/{dataMaskingRuleName}"].put.responses["200"])
```

Before:
```
ERROR (RESPONSE_VALIDATION_ERROR/OAV108/ExampleModelViolation): Invalid body: Value failed JSON Schema validation.
Scenario: Create/Update data masking rule for default min
Details: [                                                                             
  {                                                                                    
    "code": "OBJECT_ADDITIONAL_PROPERTIES",                                            
    "params": [                                                                        
      [                                                                                
        "location"                                                                     
      ]                                                                                
    ],                                                                                 
    "message": "Additional properties not allowed: location",                          
    "path": [],                                                                        
    "description": "An update request for an Azure SQL Database server."               
  }                                                                                    
]                                                                                      
More info: https://github.com/Azure/oav#oav108-response_validation_error
  - file:///c:/work/azure-rest-api-specs/specification/sql/resource-manager/Microsoft.Sql/2014-04-01/dataMasking.json:157:20 ($.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/databases/{databaseName}/dataMaskingPolicies/{dataMaskingPolicyName}/rules/{dataMaskingRuleName}"].put.responses["200"])
```

### DIFF
--- a/lib/autorestPlugin/pluginHost.js
+++ b/lib/autorestPlugin/pluginHost.js
@@ -72,7 +72,9 @@ exports.openApiValidationExample = function openApiValidationExample(swagger, sw
               throw new Error("Model Validator: Path to x-ms-examples not found.");
             }
             //console.error(JSON.stringify(scenarioItem, null, 2));
-            var result = new FormattedOutput("verbose", scenarioItem, scenario, [modelValidationCategory], "Model validator found issue (see details).", [{ document: swaggerFileName, Position: { path: xmsexPath } }])
+            var result = new FormattedOutput("verbose", scenarioItem, scenario, [modelValidationCategory],
+              "Model validator found issue (see details).",
+              [{ document: swaggerFileName, Position: { path: xmsexPath } }])
             formattedResult.push(result);
 
             // request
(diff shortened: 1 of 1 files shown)


olydis:
CC @veronicagg @amarzavery

amarzavery:
The change lgtm. However, I shall let @veronicagg decide as she knows more about this piece.

veronicagg:
@olydis For oav if the details are not included in the message, a person reading the output doesn't get enough information on what the problem is, so in my opinion they are needed details. Yes, it's more verbose that we may want, but for this tool needed. What's the issue with verbosity in this case? The tool would be called on demand when the switch is used and the info we output needs to be actionable.

olydis:
@veronicagg Fair, then I'll work on a solution that makes the output message actionable without containing JSON. It would be significantly more readable in the console. In VSCode, I'm not sure whether a 15 line message will show up the right way in the error window. I just think we can do significantly better here quality-wise. 😉 Seems to me like `description`, `path`, `params` and `code` are completely pointless in the above example's details (given the message), so there is just too much of noise for a message that's supposed to be human-readable.

veronicagg:
@olydis sure, feel free to. I tried this before in VS code (when I had it working with 1.x of autorest), and that's one of the reasons I added the new line, the first part of the message was readable and on hover we could see the entire content. Please be aware that the content of the details can vary a bunch, depending on the errors found by the validation, so it needs to adapt to that.

amarzavery:
closing this PR. Please sync with Veronica and open if needed.

PR #177, opened 2017-10-24
Disable resolving discriminators for semantic validation of a 2.0 openapi spec.

amarzavery:


veronicagg:
Issue found via https://github.com/Azure/azure-rest-api-specs/pull/1903

ChangeLog.md
```diff
@@ -1,3 +1,6 @@
+### 10/24/2017 0.4.17
+- Disable resolving discriminators while performing semantic validation for an open api specification that confirms to 2.0 version.
```
veronicagg: conforms?

lib/validators/specValidator.js
```diff
@@ -38,6 +38,8 @@ class SpecValidator {
    * 
    * @param {object} [options.shouldResolveAllOf] Should allOf references be resolved? Default: true
    * 
+   * @param {object} [options.shouldResolveDiscriminator] Should discriminator be resolved? Default: true
```
veronicagg: was this not an option in the currently committed code?

lib/validators/specValidator.js
```diff
@@ -74,6 +76,12 @@ class SpecValidator {
     if (options.shouldResolvePureObjects === null || options.shouldResolvePureObjects === undefined) {
       options.shouldResolvePureObjects = true;
     }
+    if (options.shouldResolveDiscriminator === null || options.shouldResolveDiscriminator === undefined) {
```
veronicagg: if shouldResolveDiscriminator is passed as "true" in options, should we also make sure to set "shouldResolveAllOf" as true?

veronicagg:
Also, CI is currently failing.

lib/validators/specValidator.js
amarzavery: Yup. I am doing that below in lines 81 and 82

veronicagg: Yes, but isn't that happening only if "shouldResolveDiscriminator" is null or undefined? how about if someone set it to true?

lib/validators/specValidator.js
amarzavery: I missed documenting it in the specValidator.js class. Actually specValidator simply passes these options to specResolver. So in the new commit, I have removed those options from specValidator as we are checking and validating those in the specResolver cosntructor. Hence it makes no sense to do it twice.

ChangeLog.md
amarzavery: thanks. changed it.

lib/validators/specValidator.js
amarzavery: Got it. Have changed it now. Please check the new commit.

lib/validators/specValidator.js
```diff
@@ -64,23 +64,6 @@ class SpecValidator {
...
-      options.shouldSetAdditionalPropertiesFalse = true;
-    }
-    if (options.shouldResolvePureObjects === null || options.shouldResolvePureObjects === undefined) {
-      options.shouldResolvePureObjects = true;
-    }
-    if (options.shouldResolveDiscriminator === null || options.shouldResolveDiscriminator === undefined) {
-      // Resolving allOf is a neccessary precondition for resolving discriminators. Hence hard setting this to true
-      options.shouldResolveAllOf = true;
-      options.shouldResolveDiscriminator = true;
-    }
```
amarzavery: removing the above conditions as they are being validated in the specResolver constructor. We are not exposing the resolver. It is accessed by the validator. SpecValidator is a pass thru.

.vscode/launch.json
```diff
@@ -71,7 +71,7 @@
       "cwd": "${workspaceRoot}",
       "stopOnEntry": false,
       "args": [
-        "test/liveValidatorTests.js"
+        "test/modelValidatorTests.js"
```
veronicagg: is this intentional, should it be one or the other?

amarzavery: It could be any, doesnt matter.

PR #178, opened 2017-11-07
Allow custom branch for cloned repo

vladbarosan:


lib/util/utils.js
```diff
@@ -421,15 +421,18 @@ exports.getProvider = function getProvider(path) {
...
+ * @param {string} url of the repository to be cloned.
  *                 Example "https://github.com/Azure/azure-rest-api-specs.git" or
  *                         "git@github.com:Azure/azure-rest-api-specs.git".
- * 
- * @param {string} path where to clone the repository. 
+ *
+ * @param {string} [branch] to be cloned instead of the default branch.
  */
-exports.gitClone = function gitClone(url, directory) {
+exports.gitClone = function gitClone(directory, url, branch) {
```
amarzavery: This will break Huiqing's deployments as the method signature has changed. What is the reason to make directory as the first parameter? I understand that you need to add support for branching
How about adding branch as the last parameter to the existing method signature `function gitClone(url, directory, branch) {` ?
This won't break Huiqing's existing scripts.

lib/validators/liveValidator.js
```diff
@@ -89,7 +91,7 @@ class LiveValidator {
 
     // Clone github repository if required
     if (self.options.git.shouldClone) {
-      utils.gitClone(self.options.git.url, self.options.directory);
+      utils.gitClone(self.options.directory, self.options.git.url, self.options.git.branch);
```
amarzavery: `utils.gitClone(self.options.git.url, self.options.directory, self.options.git.branch)`

PR #180, opened 2017-11-08
provide baseUrl by looking at the host and the basePath property

amarzavery:


lib/validators/specValidator.js
```diff
@@ -568,7 +568,21 @@ class SpecValidator {
...
     let options = {};
-
+    if (operation.pathObject && operation.pathObject.api && operation.pathObject.api.host) {
+      let scheme = 'https';
+      let basePath = '';
+      let host = operation.pathObject.api.host;
+      if (host.endsWith('/'))
+        host = host.slice(0, host.length - 1);
+      if (operation.pathObject.api.schemes && !operation.pathObject.api.schemes.some((item) => { return item && item.toLowerCase() === 'https'; }))
+        scheme = operation.pathObject.api.schemes[0];
```
veronicagg: if this array has multiple, how do we know we should pick the first?

lib/validators/specValidator.js
```diff
@@ -568,7 +568,21 @@ class SpecValidator {
...
+      if (operation.pathObject.api.basePath)
+        basePath = operation.pathObject.api.basePath;
+      if (!basePath.startsWith('/'))
+        basePath = `/${basePath}`;
+      let baseUrl = `${scheme}://${host}${basePath}`;
```
veronicagg: does it matter if this url ends with "/" or not?, just checking, I've seen cases in the specs where the "basePath" is defined as "/" only.

amarzavery: so if the basePath does not start with "/" only then we add forward slash.

lib/validators/specValidator.js
amarzavery: i am just picking the first one only if the array does not contain 'https'. In that scenario we any ways do not have an option.

PR #184, opened 2017-11-20
Add support for x-ms-parameterized-host

amarzavery:


amarzavery:
Changes in https://github.com/amarzavery/sway/pull/14 need to be merged for the CI to be green.

lib/validators/specResolver.js
```diff
@@ -443,6 +455,43 @@ class SpecResolver {
...
+   * ModelValidation:
+   * This step should only be performed for model validation as we need to
+   * make sure that the examples contain correct values for parameters 
+   * defined in 'x-ms-parameterized-host'.hostTemplate. Moreover, they are a 
+   * part of the baseUrl.
+   * 
+   * SemanticValidation:
+   * This step should not be performed for semantic validation, othwerise there will
+   * be a mismatch between the number of path parameters provided in the operation
+   * definition and the number pf parameters actually present in the path template.
```
veronicagg: typo: "pf" -> "of"

test/modelValidation/swaggers/specification/parameterizedhost/face.json
```diff
@@ -0,0 +1,1182 @@
+{
```
veronicagg: should we simplify the spec even more? just to leave the cases you're checking for parameterized host and have simple models that don't require that many definitions, seems a bit unnecessary for the test case?

test/modelValidatorTests.js
```diff
@@ -123,4 +123,18 @@ describe('Model Validation', function () {
       });
     });
   });
+
+  describe('for parameters in x-ms-parameterized-host', function () {
```
veronicagg: should we also have a test on semantic validation, to make sure we're not breaking it, particularly since we have the caveat that we shouldn't be applying the changes to semantic validation?

test/modelValidation/swaggers/specification/parameterizedhost/face.json
amarzavery: I understand, but that is time consuming for now. I think we should leave it as-is.
I have added a test for semantic validation

test/modelValidatorTests.js
amarzavery: added a semantic validation test as well.

Issue #155, opened 2017-08-22
Suppress INVALID_TYPE errors when the property is set to null value in the example


veronicagg:
Opening this issue to address https://github.com/Azure/azure-rest-api-specs/issues/798 in OAV tool. 
Since it's a swagger limitation to specify allowing null values, let's allow null to match any type when a property comes back with this value in a response. This way model validation should not report errors and avoids noise (we have seen a lot of case where the Azure service behaves this way, and no cases where a fix is expected on the service side).

veronicagg:
Let's check a few examples to see if the service is actually returning null or the example is wrong and should actually not have the property with null value in it.

veronicagg:
https://github.com/Azure/oav/issues/158

vishrutshah:
@amarzavery As we discussed this offline, on some lines of how should we tackle it right instead of going to the route or ignoring this, could you please add some context. Thanks!

vishrutshah:
un-assigning as my focus would on https://github.com/Azure/oav/issues/142 for next sprint.

amarzavery:
- As we know that swagger spec is a sub set of json schema. For example json schema supports oneOf, anyOf, allOf. But swagger 2.0 only supports allOf. 
- In the same way, swagger 2.0 says that the type of the schema (i.e) type of a model, type of a model property, type of a parameter can only be a string. I mean, we can say `"type": "number"`, `"type": "array"` i.e. the value for the type property can only be of type string. However, JSON schema says that "type" can be an array/combination of basic json types  `"type": ["array", "object", "boolean", "string", "number", null]` . Notice the null without quotes as a valid type.
- So we can have an option called **`options.resolveNullableTypes`** in the specResolver.js class. When set to **true** we will convert the "type" from string to an array comprising of the existing value of that type and null. Now when we give such a schema to z-schema to validate the model it should not complain.
- This will make it easy for us to not resolve nullable types for semantic validation (similar to what we do for resolving discriminators (replace with oneOf) today).
  - We can expose this via `oav resolve-spec` command with a new switch **-n** which resolve nullable types only when specified.
 - 

NOTE: We should not do this for all the model properties. This should be done only when
 -  the property is optional OR
 - the property has the extension ("x-ms-nullable": true). When this extension is applied, it does not matter whether the property is required or not. The spec author is clearly saying that he/she wants this property to accept null as a valid value and we should honor that (We have such an example in resource deployment swagger spec).

In this way we can handle the issue in a better way.

[...]

PR #185, opened 2017-11-28
Allow null types

veronicagg:
Addressing https://github.com/Azure/oav/issues/155
@amarzavery please take  a look when you get a chance, I'll be adding tests, but wanted to send a PR with the modified code, since we talked about this. 
I changed the condition on when nullable applies a bit, so want to know if you agree or we should be doing it differently. 
Basically, I noticed that we have some specs that specifically set x-nullable to false, in which case I think that even if the property is not required, we should probably honor that setting. 
Logic is now that if x-nullable is not undefined then we honor the value, and if the property is undefined, then we look at whether the property is required or not. Thoughts?


amarzavery:
@veronicagg - Please resolve conflicts with upstream master

lib/util/utils.js
```diff
@@ -580,6 +580,39 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
   return model;
 }
 
+exports.allowNullType = function allowNullType(entity, isPropRequired) {
```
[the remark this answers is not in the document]
amarzavery: The valueType could be a primitive type or an inline object with properties or a reference to another object (composite type) or an array type.
- If it is a reference to another type then it is fine.
- If it is an inline object with properties then we need to process it // allowNullableTypes()
- If it is a primitive type like string, number, boolean then we need to process it // allowNullTypes()
- If it is an array then we need to process it

veronicagg:
@amarzavery I've updated the code to account for the cases we've talked about and added tests, let me know what you think.
I see the CI is failing for other tests, I need to look into that, I wonder if my merge didn't work well, as I see something failing for x-ms-parameterized-host.

amarzavery:
@veronicagg - Please set
`options.shouldResolveNullableTypes = false` for semantic validation over [here](https://github.com/Azure/oav/blob/master/lib/validate.js#L68)

lib/commands/resolve-spec.js
```diff
@@ -53,6 +53,12 @@ exports.builder = {
     boolean: true,
     default: false
   },
+  n: {
+    alias: 'nullable',
+    describe: 'Should nullable be resolved?',
```
amarzavery: Should nullable **types** be resolved?

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
+    if (typeof entity['x-nullable'] === 'string') {
+      if (entity['x-nullable'].toLowerCase() === 'false') {
+        entity['x-nullable'] = false;
+      } else if (entity['x-nullable'].toLowerCase() === 'true') {
+        entity['x-nullable'] = true;
+      }
+    }
+
+    let isPropNullable = entity['x-nullable'] && typeof entity['x-nullable'] === 'boolean';
+    if ((isPropNullable === undefined && !isPropRequired) || isPropNullable) {
```
amarzavery: It is worthwhile providing a comment or may be putting a True False logic table that we had drawn on the white board in the comments over here. That will be helpful to us in the future.

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
+      let savedEntity = entity
```
amarzavery: // Handling nullable parameters

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
+      if (savedEntity.in) {
+        entity.oneOf = [{ "type": entity.type }, { "type": "null" }];
+        delete entity.type
+
+      }
+      else {
```
amarzavery: as per JS code style, else should be on the same line as the if closing bracket
```
} else {
 . ..
}
```

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
+      }
+    }
+
+    let isPropNullable = entity['x-nullable'] && typeof entity['x-nullable'] === 'boolean';
+    if ((isPropNullable === undefined && !isPropRequired) || isPropNullable) {
+      let savedEntity = entity
+      if (savedEntity.in) {
+        entity.oneOf = [{ "type": entity.type }, { "type": "null" }];
+        delete entity.type
+
```
amarzavery: remove extra line

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
+      else {
+        entity = {};
+        entity.oneOf = [savedEntity, { "type": "null" }];
+      }
+    }
+  }
+
+  // if there's a $ref
+  if (entity && entity["$ref"]) {
     let isPropNullable = entity['x-nullable'] && typeof entity['x-nullable'] === 'boolean';
```
amarzavery: duplicate code, can we extract this into a function and call that function?

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
     let modelProperties = model.properties;
     for (let propName in modelProperties) {
-      let isPropRequired = model.required ? model.required.some((p) => { return p == propName; }) : false
-      if (modelProperties[propName].properties) {
+      let isPropRequired = model.required ? model.required.some((p) => { return p == propName; }) : false;
+      // process properties if present
+      if (modelProperties[propName].properties || modelProperties[propName].additionalProperties) {
         modelProperties[propName] = exports.allowNullableTypes(modelProperties[propName]);
       }
       else {
```
amarzavery: else on the same line

lib/util/utils.js
```diff
@@ -580,39 +580,133 @@ exports.relaxModelLikeEntities = function relaxModelLikeEntities(model) {
...
     }
   }
-  //this needs to happen at the end, as model is modified in allowNullType to have only oneOf value with the original contents or null
-  model = exports.allowNullType(model);
+
+  if (model && model.type) {
+    if (model.type == 'array') {
+      if (model.items) {
+        // if items object contains inline properties
+        if (model.items.properties) {
```
amarzavery: what about model.items.additionalProperties?

lib/validators/specResolver.js
```diff
@@ -621,13 +621,26 @@ class SpecResolver {
     //scan definitions and properties of every model in definitions
     for (let defName in definitions) {
       let model = definitions[defName];
-      utils.allowNullableTypes(model);
+      definitions[defName] = utils.allowNullableTypes(model);
```
amarzavery: wouldn't this make the definition itself a `oneOf` of that type and null?

veronicagg:
Still pending some code review feedback, and full validation on azure-rest-api-specs repo.

amarzavery:
everything LGTM. Bump the version in package.json and also update the changelog.md

amarzavery:
This [StackOverflow link](https://stackoverflow.com/questions/47374980/schema-object-without-a-type-attribute-in-swagger-2-0) provides good explanation about requiredness `type` in a json document.

PR #189, opened 2017-12-05
Ensure discriminator property has the correct enum constant while resolving it.

amarzavery:


lib/validators/specResolver.js
```diff
@@ -657,9 +657,32 @@ class SpecResolver {
...
+      }
+      if (definitions[name].properties[discriminator]['$ref']) {
+        let obj;
+        try {
+          obj = utils.getObject(this.specInJson, definitions[name].properties[discriminator]['$ref'].slice(1));
+          if (obj && obj.type) {
+            discriminatorType = obj.type;
+          }
+        } catch (err) {
+          // do nothing, we tried to find the object and it's type. We will set it to "string" by default. It is safe to assume that properties
```
veronicagg: based on the comment the type of the discriminator should always be string? could it be something else? I'm wondering why we're doing the look-up if we want to set it to string anyway.

lib/validators/specResolver.js
```diff
@@ -657,9 +657,32 @@ class SpecResolver {
     // Adding the model name or it's discriminator value as an enum constraint with one value (constant) on property marked as discriminator
     if (definitions[name] && definitions[name].properties && definitions[name].properties[discriminator]) {
       let val = name;
+      let discriminatorType = 'string';
       if (definitions[name]['x-ms-discriminator-value']) {
         val = definitions[name]['x-ms-discriminator-value'];
       }
+      // Ensure that the property marked as a discriminator has only one enum constraint and it should be the one that we want to set as a constant.
+      if (definitions[name].properties[discriminator].enum) {
```
veronicagg: it seems that we figure out the values the discriminator should take and add them all in the polymorphic tree, therefore we'd like to ignore whatever the spec may indicate on this "type" if there was an enum defined, is this correct? If so, could we make a comment to indicate this as well?

lib/validators/specResolver.js
```diff
@@ -657,9 +657,32 @@ class SpecResolver {
...
+          // marked as "discriminator" will be of type "string" as it needs to refer to a model definition name which logically should be a 
+          // string as it is the key in the definitions object in the swagger spec.
+        }
+        delete definitions[name].properties[discriminator]['$ref'];
+      }
+      // Set the type if not set.
+      if (!definitions[name].properties[discriminator].type) {
+        definitions[name].properties[discriminator].type = discriminatorType;
+      }
       definitions[name].properties[discriminator].enum = [`${val}`];
```
veronicagg: was this line not already overwriting what the enum was, and setting it to one value? was the issue caused (for the reported case) due to having a reference for the type?

amarzavery: yes the issue was caused due to the discriminator referencing an enum type defined as a separate definition.

lib/validators/specResolver.js
amarzavery: yes that is correct. I will update the above comment to be more clear.

lib/validators/specResolver.js
amarzavery: I was not sure about this. Technically it will always be a string, since you would want to refer to a model name and that is a key in the definitions object, which has to be a string. However, I thought, if someone decides to do things differently where the value of the discriminator property is an enum of numbers [1, 2, 3] then we don't want to be in the business of assuming things. We would rather follow what `"type"` they have specified in the spec. If not found then we default it to "string".

What do you think?

veronicagg: Logic is ok, not sure if is allowed for the type to not be a string based on https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-13

PR #200, opened 2018-01-30
Fixed a typo in the variable name while resolving nullable types.

amarzavery:


lib/validators/specResolver.js
```diff
@@ -646,7 +646,7 @@ class SpecResolver {
       let pathObj = spec.paths[path];
       //need to handle paramaters at this level
       if (pathObj.parameters) {
-        for (let parameter in pathOpj.parameters) {
+        for (let parameter in pathObj.parameters) {
```
amarzavery: @fearthecowboy - thanks for pointing this. Will merge and publish as soon as the CI passes.

Issue #201, opened 2018-01-30
Fix linter issues


vladbarosan:
Running the linter on the `lib` folder produces a hunge number of issues among them misspelled variable names.

amarzavery:
@vladbarosan - AFAIK, this repo does not have any linter. It has the semantic and model validator. The linter resides over here  https://github.com/Azure/azure-openapi-validator. If this is indeed a linter issue then you may want to file it over there.

amarzavery:
Oh sorry.. I get this now. You are talking about `jshint`. Sorry got confused. Too many linters here.

amarzavery:
```
> jshint lib --reporter=jslint

<?xml version="1.0" encoding="utf-8"?>
<jslint>
        <file name="lib\autorestPlugin\pluginHost.js">
                <issue line="5" char="44" reason="Strings must use singlequote." evidence="Object.defineProperty(exports, &quot;__esModule&quot;, { value: true });" severity="W" />
                <issue line="6" char="23" reason="Identifier &apos;vscode_jsonrpc_1&apos; is not in camel case." evidence="const vscode_jsonrpc_1 = require(&quot;vscode-jsonrpc&quot;);" severity="W" />
                <issue line="6" char="50" reason="Strings must use singlequote." evidence="const vscode_jsonrpc_1 = require(&quot;vscode-jsonrpc&quot;);" severity="W" />
                <issue line="9" char="31" reason="Strings must use singlequote." evidence="const yaml = require(&quot;js-yaml&quot;);" severity="W" />
                <issue line="10" char="38" reason="Strings must use singlequote." evidence="const utils = require(&quot;../util/utils&quot;);" severity="W" />
                <issue line="15" char="53" reason="Strings must use singlequote." evidence="const openAPIDocUrl = &quot;https://github.com/Azure/oav&quot;;" severity="W" />
                <issue line="20" char="51" reason="Strings must use singlequote." evidence="const modelValidatorPluginName = &quot;model-validator&quot;;" severity="W" />
                <issue line="21" char="56" reason="Strings must use singlequote." evidence="const modelValidationCategory = &quot;ExampleModelViolation&quot;" severity="W" />
                <issue line="21" char="56" reason="Missing semicolon." evidence="const modelValidationCategory = &quot;ExampleModelViolation&quot;" severity="W" />
                <issue line="43" char="13" reason="Missing semicolon." evidence="          })" severity="W" />
[...]

PR #202, opened 2018-01-31
Improve casing errors

vladbarosan:
- Fix all linter issues
 - add option for case insensitive live validation path matching


vladbarosan:
Waiting on the sway PR to get merged https://github.com/amarzavery/sway/pull/20

lib/validators/specValidator.js
```diff
@@ -92,6 +92,7 @@ class SpecValidator {
         options.definition = self.specInJson;
         options.jsonRefs = {};
         options.jsonRefs.relativeBase = self.specDir;
+        options.isPathCaseSensitive = self.options.isPathCaseSensitive;
```
amarzavery: `self.options.isPathCaseSensitive;` Please document this option in the constructor documentation.

vladbarosan: Done

lib/util/validationResponse.js
```diff
@@ -84,9 +85,7 @@ exports.mapper = {
   'DUPLICATE_OPERATIONID': 'M6004',
   'MISSING_PATH_PARAMETER_DEFINITION': 'M6005',
   'EMPTY_PATH_PARAMETER_DECLARATION': 'M6006',
-  'MISSING_PATH_PARAMETER_DEFINITION': 'M6007',
   'EQUIVALENT_PATH': 'M6008',
-  'DUPLICATE_PARAMETER': 'M6009',
```
amarzavery: Why did we remove these constants from the mapper?

package.json
```diff
@@ -51,8 +51,8 @@
     "oav": "./cli.js"
   },
   "scripts": {
-    "jshint": "jshint index.js --reporter=jslint",
-    "test": "npm -s run-script jshint && mocha -t 100000",
+    "jshint": "jshint lib --reporter=jslint",
+    "test": "mocha -t 100000",
```
amarzavery: does npm run jshint by itself, now? Previously it did not do that. Hence we had to && the jshint script along with mocha.

amarzavery: Yup it does not run jshint by itself. https://travis-ci.org/Azure/oav/builds/335452396#L970
So please do the following:
`"test": "npm -s run-script jshint && mocha -t 100000",`

amarzavery:
please update the version in package.json and also update the changelog.

package.json
vladbarosan: @amarzavery  left it out because there as still a couple of linter issues which I was not sure to resolve. Let me look if I can get it clean and then I can add it back

amarzavery: Ah I see. I thought you fixed all of them. No worries, if you cannot fix all of them right now.

vladbarosan: @amarzavery  Getting there lets clean it up and keep it like that :)

any idea why we iterate here, doesn't seem we do anything different in each iteration: https://github.com/Azure/oav/blob/4459e377b6edb95d1d361b1888af186762c0d3dc/lib/xMsExampleExtractor.js#L192

amarzavery: you wont be able to replace this with a `for..of`. We are iterating over the responses dictionary in swagger for a given operation to extract examples for every response code
```json5
"responses": {
    "200": {},
    "202: {}
 . . 
}
```

lib/util/validationResponse.js
vladbarosan: @amarzavery  because those 2 are duplicates. Not sure if the mapped values are used anywhere with a hard dependency so I didn't adapt the numbers for remaining ones. Let me know if I should

vladbarosan:
Fixes #201 
Fixes #203

lib/util/validationResponse.js
amarzavery: @vishrutshah - There were some duplicates here. Do you know what is the effect of these error codes in reporting?

vladbarosan: @amarzavery  I don't thing right now they are used for anything since we don't have official documentation from them also I don't thing @vishrutshah  spends his time on Github anymore :)

PR #210, opened 2018-02-05
Support status code string for Live Validator

### LINKED ISSUE #208: dependency on @microsoft.azure/autorest-extension-base to be updated.
A package has been published for autorest-extension-base  as @microsoft.azure/autorest-extension-base -OAV is referencing that,. We should stop referencing it from github directly, and get it from npm. (and change any imports/require to "@microsoft.azure/autorest-extension-base"

### LINKED ISSUE #209: validate return codes both as code and as string
Most ( if not all) traffic from ARM is sending responses with the string value ( e.g. OK instead of 200 etc ). because of this if the response does not have a default value we get an error that the response code is invalid.

vladbarosan:
Fixes #209 
Fixes #208 
 - Add support for status code string for Live Validator
 - Update to use autorest-extension-base from official npm package

### DIFF
--- a/lib/validators/liveValidator.js
+++ b/lib/validators/liveValidator.js
@@ -14,7 +14,8 @@ const util = require('util'),
   Constants = require('../util/constants'),
   log = require('../util/logging'),
   utils = require('../util/utils'),
-  models = require('../models');
+  models = require('../models'),
+  http = require('http');
 
 /**
  * @class
@@ -391,6 +392,11 @@ class LiveValidator {
     }
     let request = requestResponseObj.liveRequest;
     let response = requestResponseObj.liveResponse;
+
+    if (http.STATUS_CODES[response.statusCode] === undefined && utils.statusCodeStringToStatusCode[response.statusCode] !== undefined) {
+      response.statusCode = utils.statusCodeStringToStatusCode[response.statusCode];
+    }
+
     if (!request.query) {
       request.query = url.parse(request.url, true).query;
     }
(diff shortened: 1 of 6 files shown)


lib/validators/liveValidator.js
```diff
@@ -391,6 +392,11 @@ class LiveValidator {
     }
     let request = requestResponseObj.liveRequest;
     let response = requestResponseObj.liveResponse;
+
+    if (http.STATUS_CODES[response.statusCode] === undefined && utils.statusCodeStringToStatusCode[response.statusCode] !== undefined) {
```
amarzavery: - Adding a comment would help in the future
- rather than explicitly checking for undefined, in this scenario it is better to check for a `falsey` value. what if the value is `null`, then the check will fail. Because this is our stupid javascript, it is better to check the presence of parent before accessing the nested property `response.statusCode`.

```javascript
// Added support to handle status code string (ex: "OK") along the status code number (ex: 200). 
if (response && !http.STATUS_CODES[response.statusCode] && utils.statusCodeStringToStatusCode[response.statusCode]) {
  response.statusCode = utils.statusCodeStringToStatusCode[response.statusCode]; 
}
```

amarzavery:
Please change the require statement at [this line](https://github.com/Azure/oav/blob/4dca30bf6399ac0d500be91a1d85934daa8119ad/lib/autorestPlugin/pluginHost.js#L13) to
 ```javascript
const extensionBase = require('@microsoft.azure/autorest-extension-base');'
```

lib/validators/liveValidator.js
vladbarosan: should be undefined when checking in the dictionary, but at least is shorter in this case.

vladbarosan:
@amarzavery feedback addressed.

PR #211, opened 2018-02-06
Fix autorest extension usage

fearthecowboy:
Hey @amarzavery ,

This fixes a critical break when using oav as an autorest extension.

I think there are still at least one or two other glitches; I'm still looking...


vladbarosan:
@fearthecowboy you want us to merge or still wait ?

fearthecowboy:
Merge this, it fixes critical failings; it's still not giving me results, with this it doesn't crash.

PR #214, opened 2018-02-08
use utils.getKeys() instead of utils.getValues()

amarzavery:


amarzavery:
This should fix the issue we are seeing for API management PR https://travis-ci.org/Azure/azure-rest-api-specs/jobs/339206328#L732

Issue #215, opened 2018-02-08
Fix bug in resolve external references


vladbarosan:
this was introduced in #202

PR #216, opened 2018-02-08
Fix resolve external references bug

vladbarosan:
Fixes https://github.com/Azure/oav/issues/215


vladbarosan:
yea seems fine now.  Tested on all the specs.

PR #221, opened 2018-02-28
Don't validate pure objects

### LINKED ISSUE #220: Primitive query parameters that are not required fail validation
When a query parameter is not required, the type is OneOf primtive type and null. 

When validating the url the query value won't get coerced properly because of this and this causes invalid type errors.

@amarzavery   Should we just not make the query parameters nullable ? Don't think it makes sense to have them as nullable.

vladbarosan:
Fixes #220

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,7 @@
+### 02/28/2018 0.4.33
+- Fix bug for optional query parameters where reporting invalid type [#220](https://github.com/Azure/oav/issues/220)
+- Don't validate pure objects.
+
 ### 02/05/2018 0.4.32
 - Fix bug for resolve external references introduced part of #202.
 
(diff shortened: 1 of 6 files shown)


mcardosos:
Cool! Something to add are tests.
There are already some tests for queries here:
https://github.com/Azure/oav/blob/master/test/modelValidation/swaggers/specification/query/test.json
But they don't take if params are required or not yet :)

vladbarosan:
@mcardosos those are for example validations where the issue wont come up since in the example its already parsed.

vladbarosan:
Apparently at least the live validator tests were never running because the path given was wrong and no swaggers were being loaded.

Also we should actually validate the expected errors as right now we just wait for exceptions. I opened the issue here #222

vladbarosan:
Changed title as the fix for the optional query parameters was made in Sway https://github.com/amarzavery/sway/pull/24

PR #223, opened 2018-03-03
Update package lock to refer new sway version

vladbarosan:


vladbarosan:
Since we didn't publish the new npm package, leaving the version unchanged.

PR #229, opened 2018-03-15
Add option to model implicit default response

### LINKED ISSUE #224: Enable option to model cloudError for responses even if not present in swaggers

### LINKED ISSUE #228: Validate default responses ?
For some reason it seems that Sway does not validate default responses.
 
https://github.com/amarzavery/sway/blob/21d2160311380134e31e252a7e1ce1268441c8c3/lib/types/operation.js#L303

vladbarosan:
Fixes #224 
Fixes #228

### DIFF
--- a/lib/util/utils.js
+++ b/lib/util/utils.js
@@ -759,42 +759,58 @@ function isPropertyRequired(propName, model) {
  */
 exports.statusCodeStringToStatusCode = lodash.invert(lodash.mapValues(http.STATUS_CODES, value => { return value.replace(/ |-/g, "").toLowerCase(); }));
 
-
 /**
- * Models an ARM cloud error.
+ * Models an ARM cloud error schema.
  */
-exports.CloudError = {
+exports.CloudErrorSchema = {
   description: "Error response describing why the operation failed.",
   schema: {
-    properties: {
-      error: {
-        properties: {
-          code: {
-            type: "string",
-            description: "An identifier for the error. Codes are invariant and are intended to be consumed programmatically."
-          },
-          message: {
-            type: "string",
-            description: "A message describing the error, intended to be suitable for display in a user interface."
-          },
-          target: {
-            type: "string",
-            description: "The target of the particular error. For example, the name of the property in error."
-          },
-          details: {
-            type: "array",
-            items: { type: "object" },
-            description: "A list of additional details about the error."
-          },
-          additionalInfo: {
-            type: "array",
-            items: { type: "object" },
-            description: "A list of additional info about an error."
-          }
-        },
-        additionalProperties: false
-      }
-    },
-    additionalProperties: false
+    "$ref": "#/definitions/CloudErrorWrapper"
   }
+};
+
+/**
+ * Models an ARM cloud error wrapper.
+ */
+exports.CloudErrorWrapper = {
+  type: "object",
+  properties: {
+    error: {
+      "$ref": "#/definitions/CloudError"
+    }
+  },
+  additionalProperties: false
+};
+
+/**
+ * Models a Cloud Error
+ */
+exports.CloudError = {
+  type: "object",
+  properties: {
+    code: {
+      type: "string",
+      description: "An identifier for the error. Codes are invariant and are intended to be consumed programmatically."
+    },
+    message: {
+      type: "string",
+      description: "A message describing the error, intended to be suitable for display in a user interface."
+    },
+    target: {
+      type: "string",
+      description: "The target of the particular error. For example, the name of the property in error."
+    },
+    details: {
+      type: "array",
+      items: { type: "object" },
+      description: "A list of additional details about the error."
+    },
+    additionalInfo: {
+      type: "array",
+      items: { type: "object" },
+      description: "A list of additional info about an error."
+    }
+  },
+  required: ["code", "message"],
+  additionalProperties: false
 };
--- a/lib/util/utils.js
+++ b/lib/util/utils.js
@@ -809,6 +809,9 @@ exports.CloudError = {
       type: "array",
       items: { type: "object" },
       description: "A list of additional info about an error."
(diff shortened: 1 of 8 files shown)


vladbarosan:
Will update with new reference to Sway after https://github.com/amarzavery/sway/pull/25 gets merged.

lib/util/utils.js
```diff
@@ -758,3 +758,43 @@ function isPropertyRequired(propName, model) {
...
  */
 exports.statusCodeStringToStatusCode = lodash.invert(lodash.mapValues(http.STATUS_CODES, value => { return value.replace(/ |-/g, "").toLowerCase(); }));
+
+
+/**
+ * Models an ARM cloud error.
+ */
+exports.CloudError = {
+  description: "Error response describing why the operation failed.",
+  schema: {
```
amarzavery: I see what you are trying to do. However if possible do not nest properties. That will make sure that when we use oneOf for nullable types we do not do anything drastic.

lib/validators/specResolver.js
```diff
@@ -590,6 +601,23 @@ class SpecResolver {
...
+   */
+  modelImplicitDefaultResponse() {
+    let self = this;
+    let spec = self.specInJson;
+    for (let pathObj of utils.getValues(spec.paths)) {
+      for (let operation of utils.getValues(pathObj)) {
+
+        // going through responses
+        if (operation.responses && !operation.responses.default) {
+          operation.responses.default = utils.CloudError;
```
amarzavery:
```javascript
if (!spec.definitions.CloudError) {
  spec.definitions.CloudErrorWrapper = utils.CloudErrorWrapper;
  spec.definitions.CloudError = utils.CloudError;
}
```

vladbarosan: @amarzavery so you prefer to define the cloud error in the definitions and then just make a ref to it ? that works I think

lib/util/utils.js
vladbarosan: not sure what you mean. Also here would prefer to not have nullable types since there is no reason for that.

lib/util/utils.js
```diff
@@ -758,3 +758,43 @@ function isPropertyRequired(propName, model) {
  * Contains the reverse mapping of http.STATUS_CODES
  */
 exports.statusCodeStringToStatusCode = lodash.invert(lodash.mapValues(http.STATUS_CODES, value => { return value.replace(/ |-/g, "").toLowerCase(); }));
+
+
+/**
+ * Models an ARM cloud error.
+ */
+exports.CloudError = {
```
amarzavery:
```javascript
exports.CloudErrorSchema = {
  description: "Error response describing why the operation failed.",
  schema: {
     "$ref": "#/definitions/CloudErrorWrapper"
   }
};

export.CloudErrorWrapper = {
  properties: {
      error: {
         "$ref": "#/definitions/CloudError"
      }
   },
  additionalProperties: false
};

exports.CloundError = {
  properties: {
     code: {
        type: "string",
        description: "An identifier for the error. Codes are invariant and are intended to be consumed programmatically."
     },
      message: {
            type: "string",
            description: "A message describing the error, intended to be suitable for display in a user interface."
       },
       target: {
            type: "string",
            description: "The target of the particular error. For example, the name of the property in error."
       },
       details: {
            type: "array",
            items: { type: "object" },
            description: "A list of additional details about the error."
       },
       additionalInfo: {
            type: "array",
            items: { type: "object" },
            description: "A list of additional info about an error."
       }
    },
     additionalProperties: false
  };
```

amarzavery: This should keep things in sync with our current design.  Do not forget `"type": "object"` in the model definitions for CloudErrorWrapper and CloudError. Now you know the consequences of missing `"type"` on a model definitions. Couple of questions:
1. CloudErrorWrapper has only one property "error". Shouldn't that be a required property?
2. Are there any required properties in `"CloudError"` model? I think at minimum there should be code and message. Otherwise what is the point of the error object.

amarzavery: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content - code and message are required.

vladbarosan: thanks for the link, I was looking in the specs repo and those were not marked as required. Btw should we live the innerError and add the new additionalInfo ? (If you saw the email from Gaurav)

amarzavery: Yes, I saw that email. I think we should have CloudErrorV1 and CloudErrorV2. Looks like some teams will be providing innerErrors for the already deployed api-versions. We can make our CloudError object a oneOf [V1, V2]. If we do that then we should make sure to not add CloudError during semantic validation, since oneOf is not supported in 2.0 version of the swagger spec.

vladbarosan: well are we sure those are required  ?either way this change is not intended to catch invalid cloud errors, is mostly to avoid noise from those and potentially catch any other non-described errors. So maybe I would just leave both innerError and additionalInfo. We can refine in the future

lib/util/utils.js
```diff
@@ -759,42 +759,58 @@ function isPropertyRequired(propName, model) {
...
+    target: {
+      type: "string",
+      description: "The target of the particular error. For example, the name of the property in error."
+    },
+    details: {
+      type: "array",
+      items: { type: "object" },
+      description: "A list of additional details about the error."
+    },
+    additionalInfo: {
```
amarzavery: wanna add innerError as well?

vladbarosan: yes just that the spec is a little confusing. It says the type is string but it seems to be an object. I'll just leave it without specifying any type.

vladbarosan: ah nevermind, it seems to be always an object

vladbarosan: done

PR #236, opened 2018-04-20
Update for sway to include new enum case mismatch error

vladbarosan:
- New error added for enum case mismatch


vladbarosan:
@amarzavery  please if everything seems fine.

Issue #235, opened 2018-04-17
Missing required header parameter not being caught by oav in model validation


amarzavery:
## Sample swagger path:

```
"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}/immutabilityPolicies/default/lock": {
        "post": {
          "tags": [
            "BlobContainers"
          ],
          "operationId": "BlobContainers_LockImmutabilityPolicy",
          "x-ms-examples": {
            "LockImmutabilityPolicy": {
              "$ref": "./examples/BlobContainersLockImmutabilityPolicy.json"
            }
          },
          "description": "Sets the ImmutabilityPolicy to Locked state. The only action allowed on a Locked policy is ExtendImmutabilityPolicy action. ETag in If-Match is required for this operation.",
          "parameters": [
            {
              "$ref": "../../../../../common/Parameters.json#/parameters/ResourceGroupNameParameter"
            },
            {
              "$ref": "#/parameters/AccountName"
            },
            {
              "$ref": "#/parameters/ContainerName"
            },
            {
              "$ref": "../../../../../common/Parameters.json#/parameters/ApiVersionParameter"
            },
            {
              "$ref": "../../../../../common/Parameters.json#/parameters/SubscriptionIdParameter"
            },
            {
              "name": "If-Match",
              "in": "header",
              "required": true,
              "type": "string",
              "description": "The entity state (ETag) version of the immutability policy to update. A value of \"*\" can be used to apply the operation only if the immutability policy already exists. If omitted, this operation will always be applied."
            }
          ],
          "responses": {
            "200": {
              "description": "OK -- Locks container ImmutabilityPolicy operation completed successfully.",
              "schema": {
                "$ref": "#/definitions/ImmutabilityPolicy"
              },
              "headers": {
                "ETag": {
                  "description": "The ETag HTTP response header. This is an opaque string. You can use it to detect whether the resource has changed between requests. In particular, you can pass the ETag to one of the If-Match or If-None-Match headers.",
                  "type": "string"
                }
              }
            }
          }
        }
      },
```

## Example missing header parameter 
```
{
  "parameters": {
    "subscriptionId": "45b60d85-fd72-427a-a708-f994d26e593e",
    "resourceGroupName": "res1782",
    "accountName": "sto7069",
    "containerName": "container6397",
    "api-version": "2018-02-01",
    "monitor": "true"
  },
[...]

PR #241, opened 2018-05-01
Extract headers

mcardosos:
Fixes https://github.com/Azure/oav/issues/235


amarzavery:
@mcardosos - Was this issue about extracting headers or validating headers? Can you make sure that we are actually validating header parameters (If we don't have a test then can you please write one). In that way we are sure that extraction and validation both work fine.

mcardosos:
About extracting headers

PR #254, opened 2018-05-30
More TypeScript

sergey-shandar:


cli.ts
```diff
@@ -1,15 +1,12 @@
 #!/usr/bin/env node
 
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
```
vladbarosan: should we keep the copyright info here ?

lib/autorestPlugin/pluginHost.ts
```diff
@@ -2,150 +2,204 @@
...
-  return autoRestApi.ReadFile(swaggerFileName).then((swaggerFile: any) => {
-    const swagger = yaml.safeLoad(swaggerFile);
-    return exports.openApiValidationExample(swagger, swaggerFileName).then(function (exampleValidationResults: any) {
-      for (const result of exampleValidationResults) {
-        autoRestApi.Message({ Channel: result.channel, Text: result.text, Details: result.details, Key: result.code, Source: result.source });
-      }
-      // console.error(JSON.stringify(exampleValidationResults, null, 2));
-    });
-  });
+async function analyzeSwagger(swaggerFileName: string, autoRestApi: extensionBase.Host)
```
vladbarosan: >analyzeSwagger [](start = 15, length = 14)

i was thinking of making all the functions arrow functions i.e. ``const analyzeSwagger = async (params) => {}

lib/autorestPlugin/pluginHost.ts
```diff
@@ -2,150 +2,204 @@
...
+export function openApiValidationExample(
+  swagger: yaml.DocumentLoadResult, swaggerFileName: string, options?: any): Promise<any[]>
+{
+  const formattedResult: FormattedOutput[] = []
+  if (!options) options = {}
+  options.consoleLogLevel = "off"
+  log.consoleLogLevel = options.consoleLogLevel
+  const specVal = new SpecValidator(swaggerFileName, swagger, options)
+  //console.error(JSON.stringify(swagger, null, 2))
   return specVal.initialize().then(function () {
```
vladbarosan: >then [](start = 30, length = 4)

you want to convert the thens() to await or leave it for another PR ?

lib/templates/httpTemplate.ts
```diff
@@ -1,67 +1,65 @@
...
       }
     }
-    return result;
+    return result
   }
 
-  getRequestBody() {
-    let body = ``;
+  getRequestBody(): string {
+    let body = ``
```
vladbarosan: >dy = `` [](start = 10, length = 7)

no ; ?

vladbarosan: I would maybe install TSlint in vscode so it detects issues like this

---
In reply to: [192216095](https://github.com/Azure/oav/pull/254#discussion_r192216095) [](ancestors = 192216095)

vladbarosan: or even add a tslint.json file ( we dont need to enable it as part of build yet) and we can run something like tslint  -c tslint.json './src/**/*.ts*'

---
In reply to: [192216349](https://github.com/Azure/oav/pull/254#discussion_r192216349) [](ancestors = 192216349,192216095)

PR #257, opened 2018-06-08
More TypeScript

sergey-shandar:


lib/util/utils.ts
```diff
@@ -11,14 +11,35 @@ import { log } from "./logging"
...
+export interface Definitions {
+  readonly [name: string]: Definition
+}
+
+/*
+export interface Doc {
+  readonly documents: Unknown
+  readonly definitions: Definitions
+}
+*/
```
amarzavery: remove this.

lib/autorestPlugin/pluginHost.ts
```diff
@@ -140,7 +160,7 @@ export async function openApiValidationExample(
           for (const responseCode of utils.getKeys(scenarioItem.responses)) {
```
vladbarosan: >     for (const responseCode of utils.getKeys(scenarioItem.responses)) { [](start = 5, length = 72)

maybe we can replace these nested for's wiht reducers ?
```
utils.getKeys(innerErrors).reduce( (acc, error) => {}, formattedResult); 
```

lib/autorestPlugin/pluginHost.ts
```diff
@@ -59,30 +73,35 @@ extension.Add(
     await Promise.all(promises)
   })
 
+export interface Options extends specValidator.Options {
+  consoleLogLevel?: Unknown
+}
+
 export async function openApiValidationExample(
```
vladbarosan: >function [](start = 13, length = 8)

arrow functions ?  ```export const openApiValidationExample = ( ... ) => {}```

vladbarosan: same for other places

---
In reply to: [194472396](https://github.com/Azure/oav/pull/257#discussion_r194472396) [](ancestors = 194472396)

lib/autorestPlugin/pluginHost.ts
vladbarosan: same for other cases

---
In reply to: [194471819](https://github.com/Azure/oav/pull/257#discussion_r194471819) [](ancestors = 194471819)

lib/commands/validate-example.ts
```diff
@@ -30,8 +30,8 @@ export function handler(argv: yargs.Arguments) {
     logFilepath: argv.f,
   }
   if (specPath.match(/.*composite.*/ig) !== null) {
```
vladbarosan: > if (specPath.match(/.*composite.*/ig) !== null) { [](start = 1, length = 50)

maybe regex.test(string) instead if we only care about if it matches ? not sure how much of a concern is here, but test is faster

vladbarosan: same in other places

---
In reply to: [194473800](https://github.com/Azure/oav/pull/257#discussion_r194473800) [](ancestors = 194473800)

lib/models/potentialOperationsResult.ts
```diff
@@ -18,9 +21,9 @@
  *
  */
 export class PotentialOperationsResult {
-  public readonly operations: any[]
-  public readonly reason: any
-  constructor(operations: any[], reason: any) {
+  public readonly operations: Operation[]
+  public readonly reason?: LiveValidationError
+  constructor(operations: Operation[], reason: undefined|LiveValidationError) {
```
vladbarosan: >Operation[] [](start = 26, length = 11)

can this be undefined ? also can we use the implicit assignement ? 
```
constructor( public operations: Operation[], ... ) 
```

lib/templates/httpTemplate.ts
```diff
@@ -55,31 +57,27 @@ export class HttpTemplate {
...
-    return body
   }
 
   protected getResponseBody(response: Response): string {
-    let body = ``
-    if (response && response.body !== null && response.body !== undefined) {
-      body = JSON.stringify(response.body)
-    }
-    return body
+    return response && response.body !== null && response.body !== undefined
```
vladbarosan: >response.body !== null && response.body !== undefined [](start = 23, length = 53)

isn't `(response && response.body)` enough ?

lib/util/utils.ts
```diff
@@ -571,13 +561,40 @@ export function isUrlEncoded(str: string): boolean {
   return str !== decodeURIComponent(str)
 }
 
+export interface Properties {
```
vladbarosan: >Properties [](start = 17, length = 10)

can't we use the Map<T> defined in this file here ?

lib/util/utils.ts
```diff
@@ -571,13 +561,40 @@ export function isUrlEncoded(str: string): boolean {
...
+  [name: string]: Model
+}
+
+export interface Ref {
+  readonly $ref: string
```
vladbarosan: >$ref [](start = 11, length = 4)

there is no problem in accessing this is a property (i.e. not as a dictionary key) ?

vladbarosan: actually seems fine 

---
In reply to: [194477251](https://github.com/Azure/oav/pull/257#discussion_r194477251) [](ancestors = 194477251)

lib/util/validationResponse.ts
```diff
@@ -3,10 +3,28 @@
...
+  validationCategory: string
+  code?: string
+  providerNamespace: Unknown
+  type: string
+  inner?: Error|Error[]
+  id?: Unknown
+  message?: string
+  jsonref?: string
+  "json-path"?: string
+}
```
vladbarosan: ahh :(

lib/util/validationResponse.ts
```diff
@@ -3,10 +3,28 @@
...
+
+interface Warning {
+  readonly code: Unknown
+}
 
 export class ValidateResponse {
 
-  private mapper = {
+  private readonly mapper: Map<string> = {
     SWAGGER_SCHEMA_VALIDATION_ERROR: "M6000",
```
vladbarosan: >SWAGGER_SCHEMA_VALIDATION_ERROR [](start = 4, length = 31)

I was wondering if we could have an enum for all the errors

vladbarosan: we could put the codes as the the enum values `SWAGGER_SCHEMA_VALIDATION_ERROR= "M6000`

---
In reply to: [194478201](https://github.com/Azure/oav/pull/257#discussion_r194478201) [](ancestors = 194478201)

lib/validators/liveValidator.ts
```diff
@@ -53,19 +40,48 @@ export interface Provider {
...
 /**
  * @class
  * Live Validator for Azure swagger APIs.
  */
 export class LiveValidator {
-  public cache: {
+  public readonly cache: {
     [provider: string]: Provider
-  }
+  } = {}
```
vladbarosan: use the Map<T> ?

lib/validators/liveValidator.ts
```diff
@@ -93,51 +109,51 @@ export class LiveValidator {
...
-    if (!Array.isArray(this.options.swaggerPaths)) {
-      const paths = typeof this.options.swaggerPaths
+    if (!Array.isArray(optionsRaw.swaggerPaths)) {
+      const paths = typeof optionsRaw.swaggerPaths
       throw new Error(
         `options.swaggerPaths must be of type "array" instead of type "${paths}".`)
     }
-    if (this.options.git === null || this.options.git === undefined) {
-      this.options.git = {
+    if (optionsRaw.git === null || optionsRaw.git === undefined) {
```
vladbarosan: >(optionsRaw.git === null || optionsRaw.git === undefined [](start = 7, length = 56)

`if(!optionsRaw.git) {}` ?

lib/validators/liveValidator.ts
```diff
@@ -231,15 +248,15 @@ export class LiveValidator {
 
               // Whitelist lookups: Look up knownTitleToResourceProviders
               // Putting the provider namespace onto operation for future use
-              if (title && (Constants.knownTitleToResourceProviders as any)[title]) {
-                operation.provider = (Constants.knownTitleToResourceProviders as any)[title]
+              if (title && (C.knownTitleToResourceProviders as any)[title]) {
+                operation.provider = (C.knownTitleToResourceProviders as any)[title]
               }
```
vladbarosan: @amzavery@microsoft.com  this list includes only Microsoft.Resources. is this logic still necessary and if so should we incldue other RP's ?

lib/util/constants.ts
```diff
@@ -1,84 +1,93 @@
...
+
+export const exampleInSpec = "example-in-spec"
+
+const BodyParameterValid = "BODY_PARAMETER_VALID"
+
+export const Errors = "Errors"
+
+export const Warnings = "Warnings"
+
+export const ErrorCodes = {
```
vladbarosan: >ErrorCodes [](start = 13, length = 10)

would be good to have these somehow as enums

types/sway.d.ts
```diff
@@ -3,17 +3,74 @@
 
 declare module "sway" {
   interface Options {
```
vladbarosan: @amzavery@microsoft.com  maybe we can add these to sway itself

vladbarosan:
@sergey-shandar you think after we merge this we can have a release ? I might need to make some changes and will need to release afterwards

lib/autorestPlugin/pluginHost.ts
sergey-shandar: I agree that we need to use less `for` cycles. As far as I see, we don't need to accumulate results, so `map` + `filter` may work better here.

lib/validators/liveValidator.ts
sergey-shandar: BTW, I need to rename `Map` because `Map<K, T>` is already used by ES6. `MapObject` ?

lib/validators/liveValidator.ts
sergey-shandar: In theory, it's slightly different. I would like to completely eliminate usage of `null`.

lib/autorestPlugin/pluginHost.ts
vladbarosan: don't we accumulate the results in `formattedResult` ? but yea if we don't accumulate anything, `map` for sure

lib/util/utils.ts
```diff
@@ -613,7 +630,7 @@ export function isPureObject(model: any): boolean {
  * @returns {object} entity - The transformed entity if it is a pure object else the same entity is
  * returned as-is.
  */
-export function relaxEntityType(entity: any, isRequired?: boolean) {
+export function relaxEntityType(entity: Model, isRequired?: Unknown): Model {
```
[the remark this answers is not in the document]
sergey-shandar: `specResolver` calls this function with `param.required` as this parameter. The type of this value is `Unknown[]|false`.

PR #261, opened 2018-06-18
Fixing TS types.

sergey-shandar:


lib/validators/specValidator.ts
```diff
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+/* tslint:disable-next-line:no-reference */
+/// <reference path="./../../types/sway.d.ts" />
+
```
vladbarosan: I presume the definite solution is to move these types to sway itself. Can you create an issue for this please ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.4.44",
+  "version": "0.4.48",
```
vladbarosan: >48 [](start = 18, length = 2)

maybe version 0.4.45

lib/validators/specValidator.ts
sergey-shandar: https://github.com/Azure/oav/issues/262

Issue #234, opened 2018-04-12
Fix model validation issues for Search dataplane swagger spec


amarzavery:
- Please take a look at this PR https://github.com/Azure/azure-rest-api-specs/pull/2843.
- We need to do a better job at reporting errors.

It seems there is a bug in the regex validation for paths since path contains special characters like "(" and ")" which then causes the tool to incorrectly report path parameters that were provided in the example.

Another bug seems to be around validating polymorphic types. 

Please take a look at the CI job and also follow all the pain points pointed by Bruce Johnston in the PR conversation.

mcardosos:
Did not find issues with polymorphism.
However found issues with `parametrized-host`. Open issue here https://github.com/Azure/oav/issues/240

mcardosos:
Paths issue was fixed in https://github.com/Azure/oav/pull/238

sarangan12:
Part 2 - Fixed
Part 3 - https://github.com/Azure/oav/issues/248 . needs to be fixed

Issue #248, opened 2018-05-14
Data is valid against more than one schema from \'oneOf\


sarangan12:
Repro Steps:

node cli.js validate-example https://github.com/Azure/azure-rest-api-specs/blob/master/specification/search/data-plane/Microsoft.Azure.Search.Service/preview/2017-11-11/searchservice.json

Sample Error:
```
......
code: 'ONE_OF_MISSING',
            params: [],
            message: 'Data does not match any schemas from \'oneOf\'',
            path: [ 'fields', '1' ],
            inner: 
             [ { code: 'ONE_OF_MULTIPLE',
                 params: [],
                 message: 'Data is valid against more than one schema from \'oneOf\'',
                 path: [ 'fields', '1', 'analyzer' ] },
               { code: 'INVALID_TYPE',
                 params: [ 'null', 'object' ],
                 message: 'Expected type null but found type object',
                 path: [ 'fields', '1' ] } ] },
          { code: 'ONE_OF_MISSING',
.........
```

Related to https://github.com/Azure/oav/issues/234

PR #263, opened 2018-06-19
Fixing multiple `oneOf`s

sergey-shandar:
Issue #248

### DIFF
--- a/lib/util/utils.ts
+++ b/lib/util/utils.ts
@@ -559,11 +559,12 @@ export interface Model {
   additionalProperties?: Model|false
   "x-nullable"?: Unknown
   in?: Unknown
-  oneOf?: Unknown
+  oneOf?: Model[]
   $ref?: string
   required?: Unknown[]|false
   schema?: Model
   allOf?: Ref[]
+  anyOf?: Model[]
   description?: Unknown
   discriminator?: string
   "x-ms-discriminator-value"?: string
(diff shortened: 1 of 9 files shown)


vladbarosan:
also with this , fixes https://github.com/Azure/oav/issues/234 .

lib/util/utils.ts
```diff
@@ -698,8 +699,9 @@ export function allowNullType(entity: Model, isPropRequired?: boolean|{}): Model
     && entity.$ref
     && shouldAcceptNullValue(entity["x-nullable"], isPropRequired)) {
     const savedEntity = entity
-    entity = {}
-    entity.oneOf = [savedEntity, { type: "null" }]
+    entity = {
+      anyOf: [savedEntity, { type: "null" }]
```
vladbarosan: >anyOf [](start = 6, length = 5)

is this the only place where we used to make a oneOf between the type and null ?

sergey-shandar: The one that I'm aware of that cause problems. It looks like other cases where we use `oneOf` should be good.

vladbarosan: also I was wondering, are we ready to enable tslint on the project 

---
In reply to: [196611051](https://github.com/Azure/oav/pull/263#discussion_r196611051) [](ancestors = 196611051)

sergey-shandar: @vladbarosan tslint is enabled already

PR #265, opened 2018-06-21
Resolve nested properties

sergey-shandar:


sergey-shandar:
- before https://gist.github.com/sergey-shandar/6b83e99b1bb523896f214188227b8054/d0f173e0e6581edf9d99d822daa3040d0bbe1587
- after https://gist.github.com/sergey-shandar/6b83e99b1bb523896f214188227b8054/440bfa76a7dbcbb358241c7246dda015635543aa

lib/util/mapObject.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+export interface MapObject<T> {
+  [name: string]: T
+}
+
+export function objectMap<T>(src: MapObject<T>, f: (v: T, name: string) => T): MapObject<T> {
```
vladbarosan: >function [](start = 7, length = 8)

dont we want to have arrow functions wherever we can ? or afraid to break any behavior ?

lib/util/mapObject.ts
```diff
@@ -0,0 +1,14 @@
...
+  const result: MapObject<T> = {}
+  for (const name in src) {
```
vladbarosan: >name in src) [](start = 13, length = 12)

is this the only reason we disabled forin rule ? cant we use Object.keys here  ?

vladbarosan: with `for of` on it I mean

---
In reply to: [197878939](https://github.com/Azure/oav/pull/265#discussion_r197878939) [](ancestors = 197878939)

lib/validators/specResolver.ts
```diff
@@ -167,85 +121,83 @@ export class SpecResolver {
...
+    options.shouldResolveRelativePaths = defaultIfUndefinedOrNull(
+      options.shouldResolveRelativePaths, true)
+
+    options.shouldResolveXmsExamples = defaultIfUndefinedOrNull(
+      options.shouldResolveXmsExamples, true)
+
     if (options.shouldResolveAllOf === null || options.shouldResolveAllOf === undefined) {
       if (!_.isUndefined(specInJson.definitions)) {
         options.shouldResolveAllOf = true
       }
```
vladbarosan: cant we pass this with a ternary operator to `defaultUndefinedOrNull` ?

lib/validators/specResolver.ts
```diff
@@ -281,7 +233,7 @@ export class SpecResolver {
    *
    * @return {object} doc fully resolved json document
```
vladbarosan: >return {object} doc fully resolved json document [](start = 6, length = 48)

I guess this comment is out of date ?

lib/validators/specResolver.ts
```diff
@@ -504,9 +456,9 @@ export class SpecResolver {
    *
    * @param {object} child object to be merged. Example: "Storage".
    *
-   * @return {object} returns the merged child oject
+   * @return {object} returns the merged child object
    */
-  private mergeParentAllOfInChild(parent: Model, child: Model) {
+  private mergeParentAllOfInChild(parent: JsonModel, child: JsonModel) {
```
vladbarosan: >Model) { [](start = 64, length = 8)

return type ?

lib/validators/specResolver.ts
```diff
@@ -536,10 +488,10 @@ export class SpecResolver {
   /**
    * Deletes all the references to allOf from all the model definitions in the swagger spec.
    */
-  private async deleteReferencesToAllOf(): Promise<void> {
+  private deleteReferencesToAllOf(): void {
     const self = this
     const spec = self.specInJson
```
vladbarosan: remove self  = this ?

lib/validators/specResolver.ts
```diff
@@ -990,9 +932,9 @@ export class SpecResolver {
   private buildOneOfReferences(rootNode: PolymorphicTree): Set<PolymorphicTree> {
     let result = new Set()
     result.add({ $ref: `#/definitions/${rootNode.name}` })
-    for (const entry of rootNode.children.entries()) {
-      if (entry[1]) {
-        result = new Set([...result, ...this.buildOneOfReferences(entry[1])])
+    for (const en of rootNode.children.entries()) {
```
vladbarosan: >en [](start = 15, length = 2)

can we use `for ( [enName, enObj]...)` for more clarity

vladbarosan:
>     "@microsoft.azure/autorest-extension-base": "1.0.13",

note that I locked this previously to 1.0.13 because some types were broken. Just as FYI if you want to have a look

---
Refers to: package.json:12 in d8770ad. [](commit_id = d8770adbb2dcf014e61e792e231db4fd83d89275, deletion_comment = False)

lib/util/updateProperty.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+export function updateProperty<T, K extends keyof T>(
+  obj: T, k: K, update: (v: T[K], k: K) => T[K])
+  : void {
```
vladbarosan: wouldn't be preffered to return the value instead of doing mutations on the existing object ?

vladbarosan: also cant we just use [https://lodash.com/docs/4.17.10#update](https://lodash.com/docs/4.17.10#update) ?

---
In reply to: [197904198](https://github.com/Azure/oav/pull/265#discussion_r197904198) [](ancestors = 197904198)

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+} from "yasway"
+import { updateProperty } from "../util/updateProperty"
+import { objectMap } from "../util/mapObject"
+import { methods } from "../util/methods"
+
+export function resolveNestedDefinitions(spec: JsonSpec): JsonDefinitions {
+  const result = new Result()
+  const definitions = spec.definitions
+  if (definitions) {
+    for (const name in definitions) {
```
vladbarosan: >definitions [](start = 23, length = 11)

for [defName, def]  of Object.entries(definitions) ?

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+          result.resolveParameterArray(`operations.${operation.operationId}`, operation.parameters)
+        }
+      }
+      ++i
+    }
+  }
+  return result.extra
+}
+
+class Result {
```
vladbarosan: >Result [](start = 6, length = 6)

maybe a more descriptive name for this class

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+        : undefined
+
+    updateProperty(definition, "oneOf", resolveNestedArrayProperty)
+    updateProperty(definition, "allOf", resolveNestedArrayProperty)
+    updateProperty(definition, "anyOf", resolveNestedArrayProperty)
+
+    return definition
+  }
+
+  private resolveParameter(path: string, name: string, parameter: JsonParameter): void {
```
vladbarosan: >resolveParameter [](start = 10, length = 16)

some comments on the methods ?

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+      ++i
+    }
+  }
+  return result.extra
+}
+
+class Result {
+
+  public readonly extra: JsonDefinitions = {}
+
```
vladbarosan: not sure why we need a mutable state in the result. can we just have the extra map outside of the result and we just append to it everytime we resolve a defintiion ? also it seems we are having all the definitions there not just the extra ones.

vladbarosan: something like `let allDefinitions = [...allDefinitions, resolvedDefinitions in `resolveNestedDefinitions` 

---
In reply to: [197909291](https://github.com/Azure/oav/pull/265#discussion_r197909291) [](ancestors = 197909291)

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+      case "number":
+      case "string":
+      case "boolean":
+      case "null":
+      case "file":
+        return model
+    }
+
+    const name = `${objectName}.${propertyName}`
+    const result = this.resolveDefinition(name, model)
```
vladbarosan: >resolveDefinition(name, model) [](start = 24, length = 30)

if the problem with extra is that this is recursive, can we tail recursion and return the  accumulated values

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.resolveNested(path, name, schema)
+    }
+  }
+
+  private resolveNested(objectName: string, propertyName: string, model: JsonModel): JsonModel {
+    if (model.$ref !== undefined) {
+      return model
+    }
+
+    switch (model.type) {
```
vladbarosan: >type [](start = 18, length = 4)

so if there is no type we continue ?

sergey-shandar: `"object"` or `undefined`

lib/validators/resolveNestedDefinitions.ts
sergey-shandar: We can't use tail recursion here because we traverse a tree (multiple branches).

lib/validators/specResolver.ts
sergey-shandar: Not really, the logic here is slightly different.

lib/validators/specResolver.ts
sergey-shandar: we even can use `values()` instead of `entries()`.

lib/util/updateProperty.ts
sergey-shandar: `lodash.update` doesn't provide type safety.

vladbarosan: maybe a comment on the function to explain the type restrictions

---
In reply to: [197939074](https://github.com/Azure/oav/pull/265#discussion_r197939074) [](ancestors = 197939074)

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,132 @@
...
+      case "integer":
+      case "number":
+      case "string":
+      case "boolean":
+      case "null":
+      case "file":
+        return model
+    }
+
```
vladbarosan: const primTypes =[] ; if(primTypes.some(model.type)) return

vladbarosan: maybe extract in a new method like shouldResolve()

test/specResolverTests.ts
```diff
@@ -0,0 +1,23 @@
...
+            {},
+            {}
+          ]
+        }
+      }
+    }
+    const options: Options = { shouldResolveAllOf: true }
+    const resolver = new SpecResolver("./", spec, options)
+    await resolver.resolve()
+  })
```
vladbarosan: maybe some asserts here

sergey-shandar:
See new PR #268

PR #268, opened 2018-06-29
Resolve nested Schema Objects

sergey-shandar:


lib/validators/resolveNestedDefinitions.ts
```diff
@@ -0,0 +1,129 @@
...
+import { arrayMap } from "../util/array"
+import { Tracked, tracked } from "../util/tracked";
+import * as uuid from "uuid"
+
+export function resolveNestedDefinitions(spec: SwaggerObject): SwaggerObject {
+
+  const newDefinitions: DefinitionsObject = {}
+
+  // a function to resolve nested schema objects
+  function resolveNestedSchemaObject(schemaObjectTracked: Tracked<SchemaObject>) {
```
vladbarosan: >function res [](start = 2, length = 12)

arrow function for this one ?

vladbarosan: actually why nested instead of outside and  non exported ?

---
In reply to: [199270718](https://github.com/Azure/oav/pull/268#discussion_r199270718) [](ancestors = 199270718)

vladbarosan: same for the other ones

---
In reply to: [199270922](https://github.com/Azure/oav/pull/268#discussion_r199270922) [](ancestors = 199270922,199270718)

Issue #245, opened 2018-05-11
Support serialized errors


vladbarosan:
Right now errors are reported in a nested way that makes it very hard to go through.

We should report errors as a list without the unnecessary noise coming from the nested errors. 

There might be special cases ( e.g. polymorphic types/ ONE OF MISSING) but we can work around that ).

vladbarosan:
@amarzavery  I already have a first implementation. Maybe next sprint I can have a look in improving it and adding it as an opt in feature to oav.

vladbarosan:
This is partially solved. Remaining work is at #279

PR #269, opened 2018-07-05
Error serializer.

sergey-shandar:


vladbarosan:
Solves part of https://github.com/Azure/oav/issues/245

lib/util/array.ts
```diff
@@ -4,7 +4,7 @@
 import { objectPathAppend } from "./objectPath"
 import { Tracked, tracked } from "./tracked"
 
-export function arrayMap<T>(array: Tracked<T[]>, f: (value: Tracked<T>) => T) {
+export function arrayMap<T>(array: Tracked<T[]>, f: (value: Tracked<T>) => T): T[] {
```
vladbarosan: >[] [](start = 80, length = 2)

was this causing issues ?

sergey-shandar: No. I just prefer to have explicit function signature for exported functions.

PR #272, opened 2018-07-13
Simplified errors (using `--pretty` option)

sergey-shandar:


lib/util/baseValidationError.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { Severity } from "./severity"
+import { ValidationResultSource } from "./validationResultSource"
+
+export interface BaseValidationError {
```
vladbarosan: baseValidationError vs CommonError in error.ts ? is any of them redundant?  Otherwise can we add a comment for what is each for ?

lib/util/responseReducer.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ValidationResultSource } from "./validationResultSource"
+import { Unknown } from "./unknown"
+import { ModelValidationError } from "./modelValidationError"
+
+export interface Result {
+  isValid?: Unknown
+  error?: ModelValidationError /* {
+    readonly code: Unknown
+    readonly innerErrors?: ModelValidationError[]
+  }*/
```
vladbarosan: can we remove ?

lib/util/scenarioReducer.ts
```diff
@@ -0,0 +1,115 @@
...
+  scenarios?: {
+    [key in string]?: Scenario
+  }
+  request?: Result
+  responses?: {
+    [name: string]: Result
+  }
+  error?: Unknown
+}
+*/
```
vladbarosan: also can we remove ?

lib/util/utils.ts
```diff
@@ -171,12 +171,14 @@ export function makeRequest(options: Options): Promise<SwaggerObject> {
  *
  * @return A chain of resolved or rejected promises
  */
-export async function executePromisesSequentially(
-  promiseFactories: Array<() => Promise<void>>
-): Promise<void> {
+export async function executePromisesSequentially<T>(
```
vladbarosan: >executePromisesSequentially [](start = 22, length = 27)

can we add that the return type is array of T's ?

package.json
```diff
@@ -10,12 +10,12 @@
   "license": "MIT",
   "dependencies": {
     "@microsoft.azure/autorest-extension-base": "1.0.13",
-    "@types/uuid": "^3.4.3",
     "@types/lodash": "^4.14.109",
     "@types/request": "^2.47.1",
+    "@types/uuid": "^3.4.3",
```
vladbarosan: would be good if we can remove the dependencies on the types...

sergey-shandar:
Example: image

lib/util/baseValidationError.ts
sergey-shandar: Eh, I wasn't able to use one type for now. We may revise this latter.

PR #292, opened 2018-08-13
Using a JSON parser with a source map

sergey-shandar:


cli.ts
```diff
@@ -6,7 +6,7 @@ import * as yargs from "yargs"
 import { log } from "./lib/util/logging"
 
 const defaultLogDir = log.directory
-const logFilepath = log.filepath
+// const logFilepath = log.filepath
```
vladbarosan: not used  ? then can we remove

lib/validators/modelValidator.ts
```diff
@@ -27,9 +26,9 @@ const HttpRequest = msRest.WebResource
 
 export class ModelValidator extends SpecValidator<SpecValidationResult> {
 
-  private sampleRequest: Unknown = {}
+  // private sampleRequest: unknown = {}
 
-  private sampleResponse: Unknown = {}
+  // private sampleResponse: unknown = {}
```
vladbarosan: remove ?

lib/validators/specResolver.ts
```diff
@@ -385,21 +384,21 @@ export class SpecResolver {
...
-      )
+      // TODO: doc should have a type
+      utils.setObject(doc as {}, slicedRefName, node)
+      const slicedLocalReferenceValue = parsedReference.localReference.value.slice(1)
       let referencedObj = self.visitedEntities[slicedLocalReferenceValue]
       if (!referencedObj) {
         // We get the definition/parameter from the relative file and then add it (make it local)
         // to the doc (i.e. self.specInJson) being processed.
-        referencedObj = utils.getObject(result, slicedLocalReferenceValue)
+        referencedObj = utils.getObject(result, slicedLocalReferenceValue) as SchemaObject
```
vladbarosan: >getObject [](start = 30, length = 9)

can't we type it ?

vladbarosan:
>   "version": "0.4.67",

68

---
Refers to: package-lock.json:3 in 59961d5. [](commit_id = 59961d5494a3ba475cae89e19526b59eef21df2a, deletion_comment = False)

lib/util/utils.ts
```diff
@@ -640,9 +632,9 @@ interface Entity {
   type?: DataType
   additionalProperties?: SchemaObject | boolean
   items?: SchemaObject
-  "x-nullable"?: any
+  "x-nullable"?: boolean
```
vladbarosan: >boolean [](start = 17, length = 7)

can this be ever a string ? I saw lower it its bool or string

lib/util/utils.ts
```diff
@@ -108,25 +108,31 @@ export function parseContent(
   filePath: string,
   fileContent: string
 ): SwaggerObject {
-  let result = null
   const sanitizedContent = stripBOM(fileContent)
   if (/.*\.json$/gi.test(filePath)) {
-    result = JSON.parse(sanitizedContent)
+    //  return JSON.parse(sanitizedContent)
```
vladbarosan: >//  return JSON.parse(sanitizedContent) [](start = 4, length = 39)

remove ?

lib/util/utils.ts
sergey-shandar: it can but we convert it to Boolean. We need a good Schema validator :-)

PR #293, opened 2018-08-15
Add unique array errors

vladbarosan:
- On array of inner errors, after flattening, eliminate duplicates
 - retain the paths for where those occurred


vladbarosan:
@sergey-shandar  added and fixed some lint issues from unrelated code that started popping up

PR #297, opened 2018-08-18
Passing object information using title

sergey-shandar:


lib/util/errorFileInfo.ts
```diff
@@ -1,36 +1,27 @@
...
+    try {
+      const titleObject: TitleObject|undefined = JSON.parse(title)
+      if (titleObject !== undefined) {
+        error.position = titleObject.position
+        error.url = titleObject.url
+        error.title = titleObject.title
       }
-    }
+    // tslint:disable-next-line:no-empty
+    } catch { }
```
vladbarosan: can we have a log here so we can see what are the ones without a title ?

lib/validators/specTransformer.ts
```diff
@@ -27,6 +28,28 @@ export function transform(spec: SwaggerObject): SwaggerObject {
...
+  if (info !== undefined) {
+    if (info.kind === "object") {
+      titleObject.position = {
+        line: info.position.line + 1,
+        column: info.position.column + 1,
+      }
+    }
+    titleObject.url = getFileInfo(info).url
+  }
+  model.title = JSON.stringify(titleObject)
```
vladbarosan: why stringify ? and not just leave it as an object ?

sergey-shandar: I've tried passing an object. In this case, schema validation starts failing.

vladbarosan: uhhh, i wonder why is that...ok

vladbarosan:
looks good, can we add a test for this to check that the errors will get the correct info ?

lib/validate.ts
```diff
@@ -142,11 +142,8 @@ export async function validateExamples(
     if (o.pretty) {
       /* tslint:disable-next-line:no-console no-string-literal */
       console.log(`Validating "examples" and "x-ms-examples" in  ${specPath}:\n`)
-      const errors = getErrorsFromModelValidation(
-        validator.specInJson,
-        validator.specValidationResult,
-      )
-      errorsAddFileInfo(validator.specInJson, errors)
+      const errors = getErrorsFromModelValidation(validator.specValidationResult)
+      errorsAddFileInfo(errors)
```
vladbarosan: do we need this call here ? wont it be called already inside `getErrorsFromModelValidation` which calls the `processValidationErrors`

Issue #300, opened 2018-08-22
For modelling implicit clouderror, if one is defined use that instead of the schema


vladbarosan:
`batchai/resource-manager/Microsoft.BatchAI/stable/2018-05-01/BatchAI.json` defines a `cloudError` and uses it except for some operations.

In this case, our current logic for modelling implicit default cloud error will assign "CloudErrorSchema" as a default response which wont exist. 

See the method here https://github.com/Azure/oav/blob/4fc35d48c67dbcbf13d9c9ee274301e1152714f2/lib/validators/specResolver.ts#L726

We should use the existing CloudError if it exists.

sergey-shandar:
fixed

PR #304, opened 2018-08-22
Fix for CloudError

sergey-shandar:
#300


lib/validators/modelValidator.ts
```diff
@@ -77,7 +78,7 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
       if (example === undefined) {
         throw new Error("example is undefined")
       }
-      if (toArray(keys(example as StringMap<unknown>)).length === 0) {
+      if (sm.isEmpty(toStringMap(example))) {
```
vladbarosan: >isEmpty [](start = 13, length = 7)

new feature ? :D

lib/util/utils.ts
```diff
@@ -870,26 +863,30 @@ export const statusCodeStringToStatusCode = lodash.invert(
   )
 )
 
+export const generatedCloudErrorName = generatedPrefix + "CloudError"
```
vladbarosan: >generatedPrefix [](start = 39, length = 15)

no love for interpolated strings ?

PR #306, opened 2018-08-28
Errors have optional `jsonUrl` and `jsonPosition` fields.

sergey-shandar:


lib/util/errorFileInfo.ts
```diff
@@ -25,6 +33,29 @@ const errorAddFileInfo = <T extends NodeError<T>>(error: T): void => {
...
+  /*
+  const schema = errorEx[schemaSymbol]
+  if (schema !== null && typeof schema === "object") {
+    const schemaInfo = getInfo(schema)
+    if (schemaInfo !== undefined) {
+      error.position = schemaInfo.position
+      error.url = getRootObjectInfo(schemaInfo).url
+    }
+  }
+  */
```
vladbarosan: commented code ?

lib/util/errorFileInfo.ts
```diff
@@ -25,6 +33,29 @@ const errorAddFileInfo = <T extends NodeError<T>>(error: T): void => {
       log.error(`ICE: can't parse title: ${title}`)
     }
   }
+  const errorEx = error as any as ErrorEx
```
vladbarosan: > error as any as ErrorEx [](start = 17, length = 24)

ehh

lib/util/errorFileInfo.ts
sergey-shandar: let me remove it

lib/util/errorFileInfo.ts
```diff
@@ -25,6 +33,29 @@ const errorAddFileInfo = <T extends NodeError<T>>(error: T): void => {
...
+  const json = errorEx[jsonSymbol]
+  if (json !== undefined) {
+    const jsonInfo = getInfo(json)
+    if (jsonInfo !== undefined) {
+      const path = error.path
+      error.jsonPosition = getDescendantFilePosition(
+        json,
+        path === undefined ? undefined : isArray(path) ? path : path.split("/")
+      )
+      error.jsonUrl = getRootObjectInfo(jsonInfo).url
```
vladbarosan: >getRootObjectInfo(jsonInfo).url [](start = 22, length = 31)

is this url always going to be there or do we need to check ?

lib/util/errorFileInfo.ts
```diff
@@ -25,6 +33,29 @@ const errorAddFileInfo = <T extends NodeError<T>>(error: T): void => {
       log.error(`ICE: can't parse title: ${title}`)
     }
   }
+  const errorEx = error as any as ErrorEx
+  const json = errorEx[jsonSymbol]
+  if (json !== undefined) {
+    const jsonInfo = getInfo(json)
```
vladbarosan: Maybe a test for this

lib/util/errorFileInfo.ts
sergey-shandar: It's optional so it could be `undefined`.

Issue #309, opened 2018-09-11
Error loading/parsing recording files with 0.5.8


sojain:
```
{ message: 'Errors loading/parsing recording files.',
  level: '\u001b[31merror\u001b[39m',
  [Symbol(level)]: 'error',
  [Symbol(message)]: '{"message":"Errors loading/parsing recording files.","level":"error"}' }
{ message: '{"E:\\\\work\\\\git\\\\azure-sdk-for-net\\\\src\\\\SDKs\\\\DataBox\\\\DataBox.Tests\\\\SessionRecords\\\\DataBox.Tests.Tests.JobCRUDTests\\\\TestJobCRUDOperations.json":"TypeError: entries(...) is not a function or its return value is not iterable",
```

[TestJobCRUDOperations.txt](https://github.com/Azure/oav/files/2371145/TestJobCRUDOperations.txt)

Can someone please help with the same

sergey-shandar:
@sojain could you provide more context?
- what do you run? OAV/AutoRest/etc.
- where do you run? CI/local/VSCode/etc.
- and why?

sojain:
Command run
oav extract-xmsexamples "E:\work\git\azure-rest-api-specs\azure-rest-api-specs\specification\databox\resource-manager\Microsoft.DataBox\preview\2018-01-01\databox.json" "E:\work\git\azure-sdk-for-net\src\SDKs\DataBox\DataBox.Tests\SessionRecords"

I had run it in cmd in admin mode, this is to generate examples and validate my swagger spec

sergey-shandar:
@sojain please try 0.5.10

PR #314, opened 2018-09-18
Fix for x-ms-examples extractor

sergey-shandar:
#309


documentation/source-map-design.md
```diff
@@ -1,59 +0,0 @@
-# Source Map Design
```
veronicagg: is deleting this file intentional?

lib/util/validationError.ts
```diff
@@ -15,7 +15,7 @@ export class ValidationError {
   /**
    *
    * @param name Validation Error Name
-   * @param severity The
+   * @param severity The severity
```
veronicagg: severity of the error?

test/xMsExamplesExtractor/databox.json
```diff
@@ -0,0 +1,2157 @@
+{
```
veronicagg: should the spec be simplified to target what we need to test? seems like a long spec.

veronicagg: should we have a folder "databox"  or "simple" and put all the assets corresponding to it there? the swagger and session records inside?

documentation/suppression-examples.md
```diff
@@ -0,0 +1,38 @@
+# Suppression Examples
```
veronicagg: is adding this file as part of this PR intentional?

test/xMsExamplesExtractor/databox.json
sergey-shandar: Since we had no tests for x-ms-examples-extractor, I would prefer to keep at least one set of files that gives us good coverage.

documentation/suppression-examples.md
sergey-shandar: I'm going to move the file to another branch.

veronicagg:
@sergey-shandar thanks, any comments on the question above? "what caused the issue reported by the customer? and how's this change fixing it? The error in the linked issue doesn't have much detail."

sergey-shandar:
@veronicagg the problem was caused by replacing `getEntries` by `entries` functions. There was already a varible with the same name `entries` which has type `any`. I've renamed the variable to `recordingEntries` and added export to the `entries` function.

https://github.com/Azure/oav/blob/a7c05f3e061db6ca3f74969013b1d4ab4f5baabd/lib/xMsExampleExtractor.ts#L126

In general, the cause of this problem is weak typing (using `any` type) and no test coverage.

PR #317, opened 2018-09-21
Suppression. No `where` support yet.

sergey-shandar:


lib/validators/suppressions.ts
```diff
@@ -0,0 +1,38 @@
...
+  const cmd = md.parse(readMeStr)
+  const suppressionCodeBlock = amd.getCodeBlocksAndHeadings(cmd.markDown).Suppression
+  if (suppressionCodeBlock === undefined) {
+    return undefined
+  }
+  const suppression = amd.getYamlFromNode(suppressionCodeBlock) as amd.Suppression
+  return suppression
+}
+
+const findReadMe = (dir: string): string | undefined => {
```
veronicagg: seems fine, though you may want to reuse/check what autorest does for this - https://github.com/Azure/autorest/blob/73e55bcf65181c23538a0950a1832f96517d6ffd/src/autorest-core/lib/configuration.ts#L857

lib/util/processErrors.ts
```diff
@@ -0,0 +1,69 @@
...
+  return error
+}
+
+const createErrorProcessor = <T extends NodeError<T>>(suppression: Suppression | undefined) => {
+
+  const isSuppressed = suppression === undefined ?
+    () => false :
+    (error: T): boolean =>
+      // TODO: JSONPath: https://www.npmjs.com/package/jsonpath using jp.nodes() function.
+      suppression.directive.some(item => error.code === item.suppress)
```
veronicagg: you may want to add a comment here that you're suppressing based on error codes specified in the readme file.

PR #322, opened 2018-09-25
Source Map

sergey-shandar:


veronicagg:
@sergey-shandar what does this change fix?

sergey-shandar:
@veronicagg currently, we don't ship source map. So if an error happened in OAV it shows a reference on a JavaScript generated file instead of a TypeScript source file.

veronicagg:
@sergey-shandar ah, that makes sense, thanks!

PR #324, opened 2018-09-27
Suppression support for `from` field.

sergey-shandar:


lib/util/assignOptional.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information
+
```
veronicagg: it'd be good to add a comment on why you need this function, here or in the place where you're calling it below

lib/util/processErrors.ts
```diff
@@ -35,37 +38,89 @@ const addFileInfo = <T extends NodeError<T>>(error: T): T => {
...
+  subPath.every((s, i) => mainPath[i] === s)
+
 const createErrorProcessor = <T extends NodeError<T>>(suppression: Suppression | undefined) => {
 
   const isSuppressed = suppression === undefined ?
     () => false :
-    (error: T): boolean =>
-      // TODO: JSONPath: https://www.npmjs.com/package/jsonpath using jp.nodes() function.
+    (error: T): boolean => {
+      const urlReversed = splitPathAndReverse(error.url)
```
veronicagg: maybe a good idea to a comment on the fact that the path may be partial and you're trying to see if the path in the suppression is included in the one that comes from the error.

PR #335, opened 2018-10-09
another fix.

sergey-shandar:


lib/util/utils.ts
```diff
@@ -110,28 +127,15 @@ export async function parseJson(
         "https://raw.githubusercontent.com$2$3"
       )
     }
-    const res = makeRequest({ url: specPath, errorOnNon200Response: true })
+    const res = makeRequest({ url: specPath, errorOnNon200Response: true }).then(applySuppression)
```
billytrend: why not async/await?

billytrend: This is clearer imo

```
const swaggerObject = await makeRequest({ url: specPath, errorOnNon200Response: true })
docCache[specPath]  = applySuppression(res)
return docCache[specPath] 
```

billytrend: or maybe i mean

```
const swaggerObject = await makeRequest({ url: specPath, errorOnNon200Response: true })
applySuppression(swaggerObject)
docCache[specPath]  = Promise.resolve(swaggerObject)
return swaggerObject
```

sergey-shandar: We shouldn't await (block) for uploading document from HTTPS.

billytrend: we do this already though... see line 132

veronicagg:
@sergey-shandar could you add tests to cover the cases that weren't initially covered? and surfaced this bug?

PR #336, opened 2018-10-09
Update packages.

sergey-shandar:


veronicagg:
@sergey-shandar could you update the change log as well?

PR #339, opened 2018-10-12
Source Map for generated objects.

sergey-shandar:
OAV created artificial SchemaObject. Currently, these objects have no source map. It causes that some OAV errors reported without file url / position.

The PR adds source map information to the artificial SchemaObjects.

Before

image

After fix

image


lib/validators/specResolver.ts
```diff
@@ -211,7 +210,7 @@ export class SpecResolver {
       this.specInJson = resolveNestedDefinitions(this.specInJson, this.options)
 
       // all transformations without dependencies should be moved here)
```
billytrend: redundant comment ?

package.json
```diff
@@ -10,16 +10,16 @@
...
-    "@ts-common/tuple": "0.0.0",
-    "z-schema": "^3.24.1",
+    "@ts-common/iterator": "^0.0.38",
+    "@ts-common/json": "^0.0.20",
+    "@ts-common/json-parser": "^0.2.3",
+    "@ts-common/property-set": "^0.0.8",
+    "@ts-common/source-map": "^0.2.12",
+    "@ts-common/string-map": "^0.1.1",
+    "@ts-common/tuple": "^0.0.0",
+    "@ts-common/virtual-fs": "^0.0.1",
```
billytrend: All this bumping is not nice! It means a load of code bypasses review and is terrible for readability/maintainability

lib/validators/specValidator.ts
```diff
@@ -219,9 +220,9 @@ export class SpecValidator<T extends CommonValidationResult> {
     const err: CommonError = {
       code: code.name,
       id: code.id,
-      message: message,
-      innerErrors: innerErrors ? innerErrors : undefined
+      message: message
     }
+    setMutableProperty(err, "innerErrors", innerErrors ? innerErrors : undefined)
```
billytrend: having assignment in a separate package is not ok imo! why did i have to go through 5 github pages to see what this does?! please just do assignment normally or copy + paste the code or use a legit 3rd party library like immutablejs

lib/validators/specTransformer.ts
```diff
@@ -1,49 +1,52 @@
...
-      directives: getAllDirectives(model, []),
-    }
-  model.title = JSON.stringify(titleObject)
+export const setSchemaInfo = (
+  model: SchemaObject,
+  info: SchemaObjectInfo | undefined,
+): SchemaObject => {
+  if (info !== undefined) {
+    model.title = info.title
+    setInfoFunc(model, info.infoFunc)
```
billytrend: it's so hard to tell what any of this does when i have to jump into other repos !

lib/util/utils.ts
```diff
@@ -750,6 +752,16 @@ export function allowNullType<T extends Entity>(
...
   isPropRequired?: boolean | {}
 ): T {
+
+  const info = getSchemaObjectInfo(entity)
+
+  const nullable = () => {
+    const typeNull: SchemaObject = setSchemaInfo({ type: "null" }, info)
+    const typeArray = copyInfo(entity, [entity, typeNull])
+    const newEntity: SchemaObject = setSchemaInfo({ anyOf: typeArray }, info)
+    entity = newEntity as any
```
billytrend: can we give this a type def ?

lib/util/processErrors.ts
```diff
@@ -20,9 +20,12 @@ const addFileInfo = <T extends NodeError<T>>(error: T): T => {
     try {
       const titleObject: TitleObject | undefined = JSON.parse(title)
       if (titleObject !== undefined) {
+        const path = titleObject.path
```
billytrend: prefer `const { path } = titleObject`

billytrend: in fact, overall would prefer `const { path, position, url, title} = titleObject;` and then something like `const hydratedError = { ...error, path, position, url, title }`. Overall, this function both changes the error in place and returns the error object reference which can be confusing since I might expect a function which returns the error to give me a brand new error object and not edit in place (like most built in javascript functions). It would be a lot nicer to just copy the error and move forward with this if possible. Functions with side effects like this are harder to reason about

package.json
vladbarosan: yea kind of agree with this part . we kind of need to trust that all these changes won't introduce any regressions since that code probably hasn't been reviewed and we don't know how well is tested and it doesnt have usage to be validated in the field before we use it

---
In reply to: [224907175](https://github.com/Azure/oav/pull/339#discussion_r224907175) [](ancestors = 224907175)

vladbarosan: I would suggest for those packages either to switch to libraries validated in the field ( i.e. with high usage) or aggregate all of them in one repo 

---
In reply to: [224915262](https://github.com/Azure/oav/pull/339#discussion_r224915262) [](ancestors = 224915262,224907175)

sergey-shandar: In my experience, it's other way around. It's really really hard to change anything in different `utils` folders. It's became unmanageable if we keep utilities in a project. We definitely need componentization where we can isolate and improve small pieces of code. The test coverage is also much better (usually 100%) compare to OAV. Also, we would like to share this code.

I added `^`, so after this PR, version bump is not required until breaking changes.

In the same time, I agree that I need to involve people to review utility libraries as well.

lib/validators/specValidator.ts
sergey-shandar: I would like to use good typescript utility libraries. Sadly, there are not many of them. Most of the nice libraries, such as `lazy.js` are `javascript` libraries with sometimes broken or incorrect type definitions.

lib/util/processErrors.ts
sergey-shandar: `const hydratedError = { ...error, path, position, url, title }` I'm trying hard to switch all code to immutable and not to modify existing objects but it's not subject of this PR. This function must mutate the `error` parameter.

lib/validators/specValidator.ts
billytrend: I think it would be better to improve the type definitions of libraries that are being maintained than to write our own

lib/util/processErrors.ts
billytrend: sure that's fair.. should be considered a bit of tech debt

lib/validators/specValidator.ts
sergey-shandar: @billytrend  The challenges are: 
- someone will also need to maintain this type definition because libraries are changed 
- it's really hard to understand from documentation what is the real TypeScript signature of the functions. TypeScript is limited to describe types, especially conditional types. Some libraries have functions with context dependent parameters. For example, if the first parameter is A, then the type of the second parameter should be B.

I'm ok with this if someone will do it :-) I've tried, thanks.

lib/util/processErrors.ts
sergey-shandar: @billytrend there's HUGE tech debt in OAV :-) after we moved to TypeScript.

lib/util/utils.ts
sergey-shandar: `T`

lib/validators/specTransformer.ts
sergey-shandar: I've added comment.

lib/util/utils.ts
```diff
@@ -779,12 +791,13 @@ export function allowNullType<T extends Entity>(
       const savedEntity = entity
       // handling nullable parameters
       if (savedEntity.in) {
-        entity.anyOf = [{ type: entity.type }, { type: "null" }]
+        const typeNull: SchemaObject = setSchemaInfo({ type: "null" }, info)
```
veronicagg: what problem was this causing? and how it the new code added here fixing it?

package.json
sergey-shandar: @vladbarosan we can create a package that will combine all (most) of these `ts-common` libraries.

lib/validators/suppressions.ts
```diff
@@ -5,14 +5,15 @@ import * as path from "path"
...
+import * as vfs from "@ts-common/virtual-fs"
 
 export const getSuppressions = async (specPath: string): Promise<undefined | amd.Suppression> => {
   // find readme.md
   const readMe = await amd.findReadMe(path.dirname(specPath))
   if (readMe === undefined) {
     return undefined
   }
-  const readMeStr = await amd.readFile(readMe)
+  const readMeStr = await vfs.readFile(readMe)
```
veronicagg: is vfs stuff required to fix the issue that we're seeing with paths not valid in portal? do we need to introduce this now? we haven't reviewed or validated any of the vfs code

package.json
veronicagg: a bunch of utilities were very recently added though, not sure what's meant by "there were no significant changes in utilities"

lib/validators/specValidator.ts
sergey-shandar: @billytrend here's an example. Function `reduce` which is implemented in `lazy.js` and `lodash` in some extend. The problem with this functions, that it must have to TypeScript signatues:

```ts
export const reduce = <T, A>(
    input: Iterable<T>|undefined,
    func: (a: A, b: T, i: number) => A,
    init: A,
): A
```

and

```ts
export const reduce = <T>(
    input: Iterable<T>|undefined,
    func: (a: T, b: T, i: number) => T,
): T|undefined
```

It's possible to describe two signatures for the same function in TypeScript but you are losing all compile-time type checks in this case. So in `ts-common/iterable` I have two functions `fold` and `reduce` with different signatures https://github.com/ts-common/iterator/blob/master/index.ts#L133

lib/validators/specResolver.ts
veronicagg: if there's was a question about the comment, maybe the comment should be clarified?

lib/util/utils.ts
billytrend: nice 👍

lib/validators/suppressions.ts
sergey-shandar: `virtual-fs` is a copy from `azure-openapi-markdown`. @billytrend reviewed the `azure-openapi-markdown` 

- introducing VFS in `azure-openapi-markdown`
https://github.com/ts-common/azure-openapi-markdown/pull/11/files

- move it to separate lirary
https://github.com/ts-common/azure-openapi-markdown/pull/13

`vfs` should not be a part of `azure-openapi-markdown`.

lib/util/processErrors.ts
billytrend: Yeah but I would say it's valuable to fix this tech debt as you make small changes to the project. EG this function could have been made side-effect free in this PR while you had decent context on what it was doing and in a big change later

lib/util/utils.ts
sergey-shandar: generated types should have source map (SchemaInfo). Otherwise, z-schema and yasway will not return source information.

lib/util/processErrors.ts
sergey-shandar: @billytrend too much risk for this bug fix. This function (relaxEntryType) is the first in my list:

https://github.com/Azure/oav/pull/339/files#diff-8f10dca8bc7f28f14841b4aa28b52a8eL711

lib/validators/specValidator.ts
billytrend: I don't really see how you lose type safety here but even if this is the case i think the type-safety++ is not worth the readability-- and maintainability--

sergey-shandar: @billytrend I would like to have all, readability, maintainability, consistency, and especially type safety :-) https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Functions.md#overloads

lib/validators/suppressions.ts
sergey-shandar: @veronicagg BTW, `virtual-fs` has 100% test coverage.

sergey-shandar:
working on test.

vladbarosan:
@veronicagg can you approve please after @sergey-shandar adds the test. @sergey-shandar can you also open an issue for migrating the transformations please

sergey-shandar:
Issue #338

sergey-shandar:
@vladbarosan #340

sergey-shandar:
@veronicagg the tests are added.

test/sourceMapTests.ts
```diff
@@ -0,0 +1,22 @@
...
+      // tslint:disable-next-line:max-line-length
+      "./test/modelValidation/swaggers/specification/source-map/Microsoft.ADHybridHealthService/stable/2014-01-01/ADHybridHealthService.json",
+      undefined,
+      {
+        consoleLogLevel: "off"
+      }
+    )
+    assert.strictEqual(result.length, 1)
+    const result0 = result[0]
+    assert.notStrictEqual(result0.url, undefined)
```
veronicagg: do we know what the url is? so we could validate it exactly?

PR #341, opened 2018-10-15
Provide schema information for all Schema Objects.

sergey-shandar:
We use `title` to provide source map info to yasway. We need to provide `title` for all Schema Objects.

### DIFF
--- a/test/errorInfoTests.ts
+++ b/test/errorInfoTests.ts
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { validateExamples } from "../lib/validate"
+import * as assert from "assert"
+
+describe("error info", () => {
+  it("schema object", async () => {
+    const result = await validateExamples(
+      "./test/modelValidation/swaggers/specification/errorInfo/apimusers.json",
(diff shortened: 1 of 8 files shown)


test/errorInfoTests.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { validateExamples } from "../lib/validate"
+import * as assert from "assert"
+
+describe("error info", () => {
+  it("schema object", async () => {
```
billytrend: this should probably be more descriptive, what is the behaviour being tested ?

test/errorInfoTests.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { validateExamples } from "../lib/validate"
+import * as assert from "assert"
+
+describe("error info", () => {
+  it("error info should contain url to file", async () => {
```
veronicagg: it'd be great to indicate in the test, maybe with a comment, what is special to this swagger/example combination that is helping validate the functionality, since that's generally the key to these tests.

PR #343, opened 2018-10-17
Exported functions don't catch exceptions.

sergey-shandar:
No unit-test yet.


billytrend:
This looks like a great direction from my side. Will approve as soon as test is added.

sergey-shandar:
@billytrend @veronicagg I added the unit test.

PR #348, opened 2018-11-01
No travis-CI

sergey-shandar:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.7.12",
+  "version": "0.7.13",
```
veronicagg: per our conversation, this is happening, to see it being published via the pipeline.

PR #352, opened 2018-11-15
Fix for examples in spec.

sergey-shandar:


lib/util/getErrorsFromModelValidation.ts
```diff
@@ -22,18 +22,19 @@ export function getErrorsFromModelValidation(
...
-      if (scenarios === undefined) {
-        return undefined
-      }
+      const xMsScenarios = operation["x-ms-examples"]
+      const scenario = operation["example-in-spec"]
+      const scenarios = sm.merge(
+        xMsScenarios !== undefined && xMsScenarios.scenarios !== undefined ?
+          xMsScenarios.scenarios :
+          {},
+        scenario !== undefined ? { "example-in-spec": scenario } : {}
```
vladbarosan: >scenario !== undefined ? { "example-in-spec": scenario } [](start = 8, length = 56)

nit: maybe extract as a vriable to be cleat that we are just adding this as a scenario name: scenario

lib/validators/cloudError.ts
```diff
@@ -51,21 +52,21 @@ const cloudError: SchemaObject = {
 }
 
 const cloudErrorWrapperName = `${defaultPrefix}CloudErrorWrapper`
-const cloudErrorSchemaName = `${defaultPrefix}CloudErrorSchema`
+// const cloudErrorSchemaName = `${defaultPrefix}CloudErrorSchema`
```
vladbarosan: can we delete this

lib/validators/cloudError.ts
```diff
@@ -79,26 +80,29 @@ const cloudErrorWrapper: SchemaObject = {
 }
 
 export interface ResponsesAndDefinitions {
```
vladbarosan: maybe a comment on exlpaining these

lib/validators/cloudError.ts
```diff
@@ -51,21 +52,21 @@ const cloudError: SchemaObject = {
...
 
 /**
- * Models an ARM cloud error schema.
+ * An ARM cloud error schema.
  */
 const cloudErrorSchema = {
   description: "Error response describing why the operation failed.",
-  title: `#/definitions/${cloudErrorSchemaName}`,
+  // title: `#/definitions/${cloudErrorSchemaName}`,
   schema: {
```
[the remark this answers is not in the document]
sergey-shandar: ResponseObject can't have title

PR #354, opened 2018-11-19
fix type errors

sergey-shandar:


test/modelValidatorTests.ts
```diff
@@ -57,8 +56,14 @@ describe("Model Validation", () => {
         `swagger "${specPath} with operation "${operationIds}" should report only 1 error.`
       )
 
+      if (result[0].errorDetails === undefined) {
+        throw new Error("result[0].errorDetails === undefined")
+      }
+      if (result[0].errorDetails.similarPaths === undefined) {
```
vladbarosan: will similarPaths be undefiend if there are no similar paths or just an empty array ? would prefer empty array

lib/util/baseValidationError.ts
```diff
@@ -3,16 +3,12 @@
...
-  readonly code: string
-  similarPaths: string[]
-}
-
-export interface BaseValidationError {
+export interface BaseValidationError<T extends NodeError<T>> {
   severity?: Severity
   errorCode?: string
-  errorDetails?: string | SerializedError
+  errorDetails?: T
```
vladbarosan: since we are only chanigng the types I guess the services wont need any changes ? maybe tooling if we rely on the type, please update those also and check

test/modelValidatorTests.ts
sergey-shandar: NodeError may not have `similarPaths`

lib/util/baseValidationError.ts
sergey-shandar: Sure, I will.

PR #357, opened 2018-11-26
Use `z-schema`.

sergey-shandar:


package.json
```diff
@@ -42,9 +42,9 @@
     "vscode-jsonrpc": "^3.6.2",
     "winston": "^3.0.0",
     "yargs": "^6.6.0",
-    "yasway": "^1.3.2",
+    "yasway": "^1.3.3",
     "yuml2svg": "^3.1.0",
-    "z-schema": "^3.24.1"
+    "z-schema": "^3.24.2"
```
vladbarosan: do we need a direct dependency on z-schema here ?

sergey-shandar: yes, to use jsonSymbbol and schemaSymbol (source-map).

PR #360, opened 2018-11-28
Suppress errors in suppression readme

sergey-shandar:
@vladbarosan Here's the real changes https://github.com/Azure/oav/pull/360/commits/f7be75c44f43631fd1ad917a563e8a957041c81d

Everything else is just restructure.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -32,7 +32,8 @@ export {
   ValidationResult
 } from "./lib/util/validationError"
 
-export { executePromisesSequentially, clearCache } from "./lib/util/utils"
+export { executePromisesSequentially} from "./lib/util/utils"
+export { clearCache } from "./lib/util/documents"
 
 // Classes
 export { SpecValidator } from "./lib/validators/specValidator"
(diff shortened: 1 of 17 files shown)


index.ts
```diff
@@ -32,7 +32,8 @@ export {
   ValidationResult
 } from "./lib/util/validationError"
 
-export { executePromisesSequentially, clearCache } from "./lib/util/utils"
+export { executePromisesSequentially} from "./lib/util/utils"
```
vladbarosan: why do we need to export this ?

sergey-shandar: Because we run this function in `azure-rest-api-specs` CI. Sadly, directly from `node.js`.

sergey-shandar: I mean `clearCache`. I have no idea about `executePromiseSequentially`. As soon as we get rid of API calls from portal and `azure-restapi-specs` CIs, I can throw away all of these exports.

vladbarosan: not sure I follow this seems that it should be an internal function, only exported functions should be clear API calls like validate etc, but i saw now it is exported so I guess we have to stick with it. Not sure what you mean get rid of API calls, all node packages are in some sense libraries with APIs. sometimes they have a CLI wrapper, but thats irrelevant, at base they are libraries with an API

lib/util/jsonUtils.ts
```diff
@@ -0,0 +1,121 @@
+import { Suppression, SuppressionItem } from '@azure/openapi-markdown'
```
vladbarosan: can you point in these new files what was changed beyond splitting the old one please ?

PR #362, opened 2018-11-30
Properly handle JSON errors

sergey-shandar:


lib/validators/specValidator.ts
```diff
@@ -161,15 +163,22 @@ export class SpecValidator<T extends CommonValidationResult> {
     if (this.options.shouldResolveRelativePaths) {
       docs.clearCache()
     }
+    const errors: jsonParser.ParseError[] = []
+    const reportError = (e: jsonParser.ParseError) => errors.push(e)
```
billytrend: Ah this is why you pass it as an arg?

PR #363, opened 2018-12-04
Adds retry to github request

billytrend:


lib/util/makeRequest.ts
```diff
@@ -56,6 +57,18 @@ export type Options = request.CoreOptions &
     readonly errorOnNon200Response: unknown
   }
 
+const promisifiedRequest = (options: Options) =>
+  new Promise<{ err: any; response: request.Response; responseBody: any }>(
+    (resolve, reject) =>
+      request(options, (err, response, responseBody) => {
```
vladbarosan: node-fetch instead of request ?

lib/util/makeRequest.ts
```diff
@@ -70,34 +83,47 @@ export type Options = request.CoreOptions &
...
       }
 
-      resolve(res)
-    })
-  })
-  return await promise
+      return res
+    },
+    {
+      retries: 3
```
vladbarosan: do you mind specifing the defaults explicitly ? ( for the backoff logic and whatever else )

lib/util/makeRequest.ts
```diff
@@ -70,34 +83,47 @@ export type Options = request.CoreOptions &
...
-  reportError: jsonParser.ReportError,
+  reportError: jsonParser.ReportError
 ): Promise<SwaggerObject> {
-  const promise = new Promise<SwaggerObject>((resolve, reject) => {
-    request(options, (err, response, responseBody) => {
+  return retry(
+    async (bail, retryNr) => {
+      const { err, response, responseBody } = await promisifiedRequest(options)
+
       if (err) {
```
vladbarosan: err is going to be if the request fails ? we need to handle also error return codes

package.json
```diff
@@ -22,6 +22,7 @@
     "@ts-common/virtual-fs": "^0.1.0",
```
vladbarosan: bump the patch version of the package also please

lib/util/makeRequest.ts
vladbarosan: also this needs to be async

billytrend: it doesn't havvve to be since it returns a promise, you can await on anything not just async. But I'm going to switch to node-fetch as you suggest

lib/util/makeRequest.ts
```diff
@@ -69,35 +63,42 @@ export type Options = request.CoreOptions &
  * @return {Promise} promise - A promise that resolves to the responseBody or rejects to an error.
  */
 export async function makeRequest(
-  options: Options,
-  reportError: jsonParser.ReportError,
+  url: string,
```
sergey-shandar: Please, fix all calls to this function.

lib/util/makeRequest.ts
```diff
@@ -82,45 +63,38 @@ const promisifiedRequest = (options: Options) =>
...
+    async (bail, retryNr): Promise<SwaggerObject> => {
+      try {
+        const response = await fetch(url)
+        const body = await response.text()
 
-      if (options.errorOnNon200Response && response.statusCode !== 200) {
-        const msg = `StatusCode: "${
-          response.statusCode
-        }", ResponseBody: "${responseBody}."`
+        if (response.status !== 200) {
```
vladbarosan: are we sure its going to be a 200 ? maybe we can do > 399 ?

lib/util/makeRequest.ts
```diff
@@ -82,45 +63,38 @@ const promisifiedRequest = (options: Options) =>
...
+          const msg = `StatusCode: "${
+            response.status
+          }", ResponseBody: "${body}."`
 
-        return bail(new Error(msg))
-      }
+          return bail(new Error(msg)) as any
```
vladbarosan: and why are we bailing on a bad code ?

vladbarosan: maybe its some temp issue where its failing, since its not really an api call I guess probably its fine, but I would just retry in this case also

lib/util/makeRequest.ts
billytrend: why wouldn't it be 200?

lib/util/makeRequest.ts
billytrend: hm, retry even on 404?

lib/util/makeRequest.ts
vladbarosan: could be a redirect or god knows, was airing on the safer side

lib/util/makeRequest.ts
vladbarosan: well, i would say no, but the api returns 404 for some weird reasons so your call. and if for some reason github returns 404 because the resource is not accessible for a little while retryig wont harm us

lib/util/jsonUtils.ts
```diff
@@ -90,8 +95,7 @@ export async function parseJson(
   if (doc) {
     return await doc
   }
-  // url
-  if (specPath.match(/^http.*/gi) !== null) {
+
     // If the spec path is a url starting with https://github then let us auto convert it to an
     // https://raw.githubusercontent url.
     if (specPath.startsWith("https://github")) {
```
billytrend: @sergey-shandar should i move this to vfs potentially?

PR #366, opened 2018-12-05
Remove generated CloudError

sergey-shandar:


oav.code-workspace
```diff
@@ -0,0 +1,13 @@
+{
+	"folders": [
```
vladbarosan: I presume this was added by mistake ?

ChangeLog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 12/05/2018 0.9.8
```
vladbarosan: i think we need to do 0.10.0 on this since its going to be a breaking change

oav.code-workspace
sergey-shandar: no, you can still open using either a folder or the workspace. The workspace contains a relative path to yasway which simplifies testing yasway changes for OAV.

PR #369, opened 2018-12-11
Update export

veronicagg:
Exporting NodeError to use it in OpenAPI Hub project.


billytrend:
Looks good but can you just do like `git rebase --onto origin/master HEAD~1 HEAD --autostash` and then force push? I think only the most recent commit is required?

billytrend:
oh actually, we squash and merge so it's fine don't worry

Issue #271, opened 2018-07-09
`$schema` references in examples don't have to be resolved


jianghaolu:
If an x-ms-example contains a `$schema` field in this response payload:

https://github.com/koralski/azure-rest-api-specs/blob/e3d9ca01eb41c555b03705a38e58173de2196194/specification/compute/resource-manager/Microsoft.Compute/stable/2018-04-01/examples/VirtualMachineRunCommandGet.json#L30

OAV will try to resolve the object from the link. This is not necessary in model validation.

sergey-shandar:
Fixed.

PR #374, opened 2019-01-14
OAV shouldn't fail on examples which contain "$schema" properties.

sergey-shandar:
Issue #271

### DIFF
--- a/lib/validators/specResolver.ts
+++ b/lib/validators/specResolver.ts
@@ -388,6 +388,8 @@ export class SpecResolver {
           slicedRefName.match(regex) === null)
       ) {
         // TODO: doc should have a type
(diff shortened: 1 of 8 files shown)


lib/validators/specResolver.ts
```diff
@@ -390,7 +388,7 @@ export class SpecResolver {
           slicedRefName.match(regex) === null)
       ) {
         // TODO: doc should have a type
-        utils.setObject(doc as {}, slicedRefName, result)
+        utils.setObject(doc as {}, slicedRefName, () => result)
       }
```
vladbarosan: maybe add a comment here for future reference so nobody comes and deletes it :)

PR #378, opened 2019-01-18
Handle `text-matches` in suppression.

sergey-shandar:


lib/util/processErrors.ts
```diff
@@ -55,10 +55,19 @@ const addFileInfo = <T extends NodeError<T>>(error: T): T => {
...
-const isSuppressed = <T extends NodeError<T>>(error: T): boolean =>
-  error.directives !== undefined &&
-  error.code !== undefined &&
-  error.directives[error.code] !== undefined
+const isSuppressed = <T extends NodeError<T>>({ code, directives, message }: T): boolean => {
+  if (directives === undefined || code === undefined) {
+    return false
+  }
+  const messageRegEx = directives[code]
+  if (messageRegEx === undefined || typeof messageRegEx !== "string") {
```
vladbarosan: isn't the type of messageRegEx a string ?

lib/util/jsonUtils.ts
```diff
@@ -16,13 +16,16 @@ import { log } from "./logging"
...
+  info: FilePosition | undefined,
+  item: SuppressionItem
+) => {
   if (info !== undefined) {
     if (info.directives === undefined) {
       (info as any).directives = {}
     }
-    const directives = info.directives as MutableStringMap<boolean>
-    directives[code] = true
+    const directives = info.directives as MutableStringMap<string>
```
vladbarosan: why as instead of having directives typed ?

PR #380, opened 2019-01-22
Updage z-schema

sergey-shandar:
See also issue #381


test/modelValidatorTests.ts
```diff
@@ -55,20 +55,22 @@ describe("Model Validation", () => {
       )
 
       assert(
-        result.length === 1,
+        result.length === 2,
```
vladbarosan: can we add a comment her ethat it should ne one and tracked by the 381 issue ?

PR #383, opened 2019-01-29
More errors with position and URL

sergey-shandar:


lib/validators/specValidator.ts
```diff
@@ -159,7 +162,7 @@ export class SpecValidator<T extends CommonValidationResult> {
    * Initializes the spec validator. Resolves the spec on different counts using the SpecResolver
    * and initializes the internal api validator.
    */
-  public async initialize(): Promise<Sway.SwaggerApi> {
+  public readonly initialize = async (): Promise<Sway.SwaggerApi> => {
```
vladbarosan: any advantage of this ? although of fan of arrow functions, i tend to avoid them for class instance methods for a couple of reasons:
   - harder to mock
   - breaks inheritance
  - performance goes down dramatically because the functions are not in the prototype

sergey-shandar: That's the only way (I've found) how to make the function immutable. https://github.com/ts-common/ts-common.github.io/blob/master/examples/functions.ts

sergey-shandar: Please, no inheritance :-)

vladbarosan: given the downsides, any specific reason we want to make the instance functions immutable ? What's the problem we have with that ?

ChangeLog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+### 01/29/2019 0.13.4
+
+- Provide url and file position for these errors:
+  - `RESPONSE_STATUS_CODE_NOT_IN_SPEC`,
```
vladbarosan: anything left ?

lib/validators/specValidator.ts
```diff
@@ -220,31 +223,38 @@ export class SpecValidator<T extends CommonValidationResult> {
...
+    source?: json.JsonObject
+  ): TE => {
 
-    const err: CommonError = {
+    const err: TE = {
       code: code.name,
       id: code.id,
       message: message
-    }
+    } as TE
```
vladbarosan: >as TE [](start = 6, length = 5)

this is needed ?

lib/validators/specValidator.ts
```diff
@@ -220,31 +223,38 @@ export class SpecValidator<T extends CommonValidationResult> {
...
    */
-  protected constructErrorObject<TE extends CommonError>(
+  protected readonly constructErrorObject = <TE extends CommonError>(
     code: ErrorCode,
     message: string,
     innerErrors?: null | TE[],
-    skipValidityStatusUpdate?: boolean
-  ): TE {
+    skipValidityStatusUpdate?: boolean,
+    source?: json.JsonObject
```
vladbarosan: nit: can we make a params object so we avoid passing in `undefined` ? also would make it easier to see what the params are when used since there are a lot of them

lib/validators/specValidator.ts
sergey-shandar: Sad but yes.

Issue #325, opened 2018-10-01
Two examples of same shape, oav report error for only one.


anuchandy:
Refer: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/427480126

Following two examples has the same shape.

1. https://github.com/Azure/azure-rest-api-specs/blob/76883a32f2c9aa596a7c0ceefb356637708d6ed3/specification/mysql/resource-manager/Microsoft.DBforMySQL/stable/2017-12-01/examples/ServerCreateGeoRestoreMode.json

2. https://github.com/Azure/azure-rest-api-specs/blob/76883a32f2c9aa596a7c0ceefb356637708d6ed3/specification/mysql/resource-manager/Microsoft.DBforMySQL/stable/2017-12-01/examples/ServerCreatePointInTimeRestore.json

but oav reports error for only one example, ServerCreateGeoRestoreMode.json.

```
operationId: Servers_Create
scenario: 'Create a server as a geo restore '
source: request
responseCode: ALL
severity: 0
errorCode: ONE_OF_MULTIPLE
errorDetails:
  code: ONE_OF_MULTIPLE
  params: []
  message: Data is valid against more than one schema from 'oneOf'
  path: properties
  title: /definitions/ServerForCreate/properties/properties
  description: Properties of the server.
  position:
    line: 1582
    column: 23
  url: >-
    c:\code\azure-rest-api-specs\specification\mysql\resource-manager\Microsoft.DBforMySQL\stable\2017-12-01\mysql.json

```

PR #387, opened 2019-02-06
Fix x-ms-enum as discriminator

sergey-shandar:
Issue #325

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 02/05/2019 0.13.5
+
+- Fixing `ONE_OF_MULTIPLE`.
+
 ### 01/29/2019 0.13.4
 
 - Provide url and file position for these errors:
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,7 +2,8 @@
 
 ### 02/05/2019 0.13.5
 
-- Fixing `ONE_OF_MULTIPLE`.
+- Fixing `ONE_OF_MULTIPLE`. It was caused by `x-ms-enum.modelAsString` which validates
+  discriminators as an unrestricted strings
 
 ### 01/29/2019 0.13.4
 
--- a/lib/util/jsonRefs.ts
+++ b/lib/util/jsonRefs.ts
@@ -1,0 +1,10 @@
+import * as jsonRefs from "json-refs"
+import * as sm from "@ts-common/string-map"
+
+export { UnresolvedRefDetails } from "json-refs"
+
+export const findRefs = (
+  obj: object | ReadonlyArray<unknown>,
+  options?: jsonRefs.JsonRefsOptions
+): sm.StringMap<jsonRefs.UnresolvedRefDetails> =>
+  jsonRefs.findRefs(obj, options) as sm.StringMap<jsonRefs.UnresolvedRefDetails>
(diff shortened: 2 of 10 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 02/05/2019 0.13.5
+
+- Fixing `ONE_OF_MULTIPLE`.
```
vladbarosan: a couple details on what is actually fixed on it :)

lib/validators/specResolver.ts
```diff
@@ -931,6 +933,10 @@ export class SpecResolver {
         if (d.$ref) {
           delete d.$ref
         }
+        const xMsEnum = d["x-ms-enum"]
+        if (xMsEnum !== undefined) {
+          ps.asMutable(xMsEnum).modelAsString = false
```
vladbarosan: i would say explain why we do this

lib/validators/specResolver.ts
```diff
@@ -832,7 +834,7 @@ export class SpecResolver {
    */
   private updateReferencesWithOneOf(
     subTreeMap: Map<string, PolymorphicTree>,
-    references: Array<{ readonly uri: unknown }>
+    references: StringMap<jsonRefs.UnresolvedRefDetails>
```
vladbarosan: why not use Map<string, UnresolvedRefDetails>

lib/util/jsonRefs.ts
```diff
@@ -0,0 +1,10 @@
+import * as jsonRefs from "json-refs"
+import * as sm from "@ts-common/string-map"
+
+export { UnresolvedRefDetails } from "json-refs"
+
+export const findRefs = (
```
vladbarosan: do we really need this ?

lib/validators/specResolver.ts
sergey-shandar: because it's what we receive from `json-refs`. Not `Map`.

lib/util/jsonRefs.ts
sergey-shandar: To make it strongly typed. `json-refs` has not-complete type definition https://github.com/whitlockjc/json-refs/pull/151

vladbarosan: ok so i suppose its a temporary workaround, can you refer the issue please so we can update once that is fixed

lib/validators/specResolver.ts
veronicagg: there are cases were this should be true, like on enums not part of discriminator types, how are we affecting those cases? we generally recommend modelAsString= true for specs, to avoid breaking changes in future api-versions for code generation, so I'd like to understand more of what's not working due to this value for the tool

veronicagg:
@sergey-shandar could you expand on your resolution for my question/comment: "there are cases were this should be true, like on enums not part of discriminator types, how are we affecting those cases? we generally recommend modelAsString= true for specs, to avoid breaking changes in future api-versions for code generation, so I'd like to understand more of what's not working due to this value for the tool"

sergey-shandar:
@veronicagg this fix is only applies for a specific model (internally generated) when we know exact value that we expect. https://github.com/Azure/oav/blob/d5bfd41794aa5c1fdd6b1f1a83971b5bba5c190d/lib/validators/specResolver.ts#L950

PR #389, opened 2019-02-07
fix type for polymorphic objects

sergey-shandar:


lib/validators/specResolver.ts
```diff
@@ -922,6 +922,8 @@ export class SpecResolver {
     // (constant) on property marked as discriminator
     const definition = definitions[name]
     if (definition && definition.properties) {
+      // all derived types should have `"type": "object"`.
+      definition.type = "object"
```
vladbarosan: can u please point to the documentation specifying this for future reference .
Also what happens if the definition is actually not an object ? are we going to surface that as an issue in the semantic validator ?

sergey-shandar: We use z-schema for validation, if there is no `type`, z-schema assumes that it could be anything (ex. `string`, `number`, etc). In this case, almost any value (except object) will pass validation. Which's definitely is not what we want.

lib/validators/specResolver.ts
```diff
@@ -922,8 +922,18 @@ export class SpecResolver {
...
+      // all derived types should have `"type": "object"`.
+      // otherwise it may pass validation for other types, such as `string`.
+      // see also https://github.com/Azure/oav/issues/390
+      definition.type = "object"
       const d = definition.properties[discriminator]
       if (d) {
+        const required = definition.required
+        if (!isArray(required)) {
+          definition.required = [ discriminator ]
+        } else if (find(required, v => v === discriminator) === undefined) {
```
vladbarosan: Array.find()

lib/validators/specResolver.ts
```diff
@@ -922,8 +922,18 @@ export class SpecResolver {
...
+          definition.required = Array.from(concat(required, [discriminator]))
```
vladbarosan: spread operator

PR #393, opened 2019-03-06
Remove global cache

vladbarosan:
- remove global cache on the package level. This causes issue when validating multiple files and also creates memory leaks. state should be at a class level, not module level.


billytrend:
Did you turn off caching completely?

vladbarosan:
@billytrend , not sure by completely, whatever uses a local cache will use that, there is no global cache anymore

vladbarosan:
to note that as it was mostly the global cache was not even really used as validations where using resolveRelativeReferences which would clear the cache.

PR #401, opened 2019-03-19
Change test runner to jest

billytrend:
This is to allow us to leverage jest snapshot testing.

Includes: 

* vscode launch.json jobs
* source maps
* tests take the same amount of time as before


test/wireFormatGeneratorTests.ts
```diff
@@ -1,24 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-import * as glob from "glob"
-import * as jsYaml from "js-yaml"
-import * as path from "path"
-// tslint:disable-next-line: no-implicit-dependencies
-import * as should from "should"
-
-const yamlPaths = glob.sync(path.join(__dirname, "swaggers/**/yaml/*.yml"))
```
billytrend: afaict no files match this glob so no tests were running for this file

package.json
```diff
@@ -111,7 +111,7 @@
   "scripts": {
     "tsc": "tsc",
     "tslint": "tslint --project tsconfig.json ./*.ts ./lib/**/*.ts ./test/**/*.ts ./types/**/*.ts",
-    "test": "npm run tsc && npm run tslint && nyc mocha ./dist/test/**/*.js -t 10000 --reporter mocha-junit-reporter --reporter spec",
+    "test": "npm run tsc && npm run tslint && jest --ci --reporters=default --reporters=jest-junit",
```
sergey-shandar: Does Jest have code coverage? Does it use `NYC` so we can see coverage for `.ts` files?

sergey-shandar: We also need to report to test-results.xml so Azure DevOps recognize it.

vladbarosan: it has, we use in the portal

vladbarosan: i think we have the same config as in the portal and there it gets picked up, maybe something else needs to be done if it doesnt show now

billytrend: done

billytrend: yeah it looks exactly the same as how it was being output before

test/wireFormatGeneratorTests.ts
billytrend: @vladbarosan can you confirm this is fine or tell me what files this should be run on please?

vladbarosan: @billytrend no idea what the wireFormatGenerator is @sergey-shandar  you know ?

PR #403, opened 2019-03-20
Add regression suites for validateExamples

billytrend:
Adds fast (10 mins) and slow (1 hour) regression.

fast: https://dev.azure.com/azure-sdk/public/_build?definitionId=166&_a=completed
slow: https://dev.azure.com/azure-sdk/public/_build?definitionId=163&_a=completed

The fast one runs on all latest stable specs and is intended for merge validation and the slow one is intended for a full test.

If the slow one fails, it will produce a git patch as an artefact to update snapshots locally for maximum convenience.


nschonni:
It would probably be good to add this to an `.npmignore` so it doesn't get published to NPM

sergey-shandar:
@nschonni @billytrend actually we don't use `.nmpignore` because we use `"files"` in the `package.json`.

billytrend:
@sergey-shandar how do i get the fast test to appear here: 
image

azure-pipelines.yml
```diff
@@ -1,47 +1,32 @@
...
-    testResultsFiles: '**/test-results.xml'
-    testRunTitle: 'Test results for JavaScript'
-
-- task: PublishCodeCoverageResults@1
-  inputs:
-    codeCoverageTool: Cobertura
-    summaryFileLocation: '$(System.DefaultWorkingDirectory)/**/*coverage.xml'
-    reportDirectory: '$(System.DefaultWorkingDirectory)/**/coverage'
-
-- task: Npm@1
```
sergey-shandar: Why `npm pack` is removed?

billytrend: Revert this file now

.ci/validate-examples-regression-sample.yml
```diff
@@ -0,0 +1,44 @@
+pool:
```
sergey-shandar: Could we have only one pipeline but with conditions?

vladbarosan: i think its clearer to have 2 seperate pipelines

vladbarosan: then you can look at the history easier

sergey-shandar: The problem is that you will need to create two or more Azure Dev Ops pipelines. It would be good if we keep our CI configuration in source code as much as possible.

README.md
```diff
@@ -126,6 +126,13 @@ await apiValidator.initialize() // Note that for a large number of specs this ca
 const validationResult = apiValidator.validateLiveRequestResponse(requestResponsePair)
 ```
 
+### Regression testing
+
+Output of the OAV tool has been snapshotted and committed to the repo. The regression test may be run on a sample or all of https://github.com/azure/azure-rest-api-specs. If there are changes to the snapshots the build produces a git patch file as an artifact which may be used to update the snapshots.
+
+[Fast Regression (~10mins)](https://dev.azure.com/azure-sdk/public/_build?definitionId=166&_a=completed)
```
vladbarosan: the fast one will run as a merge validation correct ?

billytrend: yes

PR #405, opened 2019-03-27
Seperate request response

vladbarosan:
- refactor live validator
- expose separate validate request and response calls.


lib/validators/liveValidator.ts
```diff
@@ -9,7 +9,7 @@ import * as msRest from "ms-rest"
 import * as os from "os"
 import * as path from "path"
 import { ParsedUrlQuery } from "querystring"
-import * as url from "url"
+import * as URL from "url"
```
sergey-shandar: It should be `url`. Even is it's an abbreviation.

lib/models/requestResponse.ts
```diff
@@ -42,10 +42,10 @@ export class RequestResponse {
   /**
    * Defines the metadata of RequestResponse
    *
-   * @returns {object} metadata of RequestResponse
+   * @returns metadata of RequestResponse
    *
    */
-  public mapper() {
+  public mapper(): object {
```
sergey-shandar: I believe, this is autogenerated file.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.14.8",
+  "version": "0.14.9",
```
sergey-shandar: Is it potentially breaking changes?

vladbarosan: well so far its not, although im thinking of making some breaking changes

lib/models/requestResponse.ts
vladbarosan: ah yea I saw the notice now

lib/validators/liveValidator.ts
vladbarosan: eh I changed it since I was getting some shadowed variable tslint errors

lib/validators/liveValidator.ts
```diff
@@ -332,29 +350,132 @@ export class LiveValidator {
...
 
+  /**
+   *  Validates live request.
+   */
+  public validateLiveRequest(
+    liveRequest: LiveRequest,
+    specOperation?: Operation
+  ): RequestValidationResult {
+    let operation = specOperation
+    if (!operation) {
```
ericwill-msft: Is there a use case for validating the request against a different operation than the one specified by the request url + method? If not we may want to remove specOperation from the parameters.

vladbarosan: well, that part was introduced mostly so that if we already have the operation, we dont need to find it again ( see usage in `validateRequestResponse`)

ericwill-msft: Couple questions then:
- Do we return the operation so it can be passed in again? Looks like we return an OperationInfo which has an OperationId but not an Operation... 
- Should we check for cases where the provided Operation doesn't match what we would have resolved if it were not provided? Maybe this is an error/warning condition?

ericwill-msft:
Do we need additional tests to cover the changes?

lib/validators/liveValidator.ts
vladbarosan: > Do we return the operation so it can be passed in again? Looks like we return an OperationInfo which has an OperationId but not an Operation...
We dont want to return the operation, we just return the validation result ( which will contain the op_id against it was executed ). usually you would need like you mentioned only the method and url, thats why we don't expose the `findSpecOperation` since except internally doesnt really makes sense to have it ( unless you test some stuff and you keep the url and method I guess and just change the payload, for that case we could expose it ).

> Should we check for cases where the provided Operation doesn't match what we would have resolved if it were not provided? Maybe this is an error/warning condition?
Wellll, i mean in that case we can even skip passing it, since the point is to skip having to do that.

vladbarosan:
> Do we need additional tests to cover the changes?
@ericwill-msft  yes. now that we have the response type ill add a couple of tests for all 3 scenarios  ( request, response, pair)

vladbarosan:
guys if there is anything else please have a look, otherwise good to go from my side, finally live validation tests in oav :)

vladbarosan:
as a follow up will look to make the inner error a more specific type

PR #408, opened 2019-04-01
Improve inner error type

vladbarosan:


test/__snapshots__/liveValidatorTests.ts.snap
```diff
@@ -41,7 +165,7 @@ Object {
       Object {
         "code": "ENUM_CASE_MISMATCH",
         "documentationUrl": "",
-        "inner": undefined,
+        "inner": Array [],
```
sergey-shandar: Please, leave it as undefined.

PR #410, opened 2019-04-05
fix exit code

sergey-shandar:


lib/commands/validate-example.ts
```diff
@@ -33,9 +34,11 @@ export async function handler(argv: yargs.Arguments): Promise<void> {
       pretty: argv.p
     }
     if (specPath.match(/.*composite.*/gi) !== null) {
-      await validate.validateExamplesInCompositeSpec(specPath, vOptions)
+      const result = await validate.validateExamplesInCompositeSpec(specPath, vOptions)
+      return iterator.flat(result).some(() => true) ? 1 : 0
     } else {
-      await validate.validateExamples(specPath, operationIds, vOptions)
+      const result = await validate.validateExamples(specPath, operationIds, vOptions)
+      return iterator.some(result, () => true) ? 1 : 0
```
veronicagg: is result here an array of errors and empty if no errors exist? if so, why not return `result.length > 0 ? 1 : 0`

lib/commands/validate-spec.ts
```diff
@@ -21,9 +22,11 @@ export async function handler(argv: yargs.Arguments): Promise<void> {
       pretty: argv.p
     }
     if (specPath.match(/.*composite.*/gi) !== null) {
-      await validate.validateCompositeSpec(specPath, vOptions)
+      const result = await validate.validateCompositeSpec(specPath, vOptions)
+      return iterator.some(result, v => !v.validityStatus) ? 1 : 0
     } else {
-      await validate.validateSpec(specPath, vOptions)
+      const result = await validate.validateSpec(specPath, vOptions)
+      return result.validityStatus ? 0 : 1
```
veronicagg: can you remind me what validityStatus is for?

lib/cliSuppressExceptions.ts
```diff
@@ -3,9 +3,9 @@
 
 import { log } from "./util/logging"
 
-export const cliSuppressExceptions = async (f: () => Promise<void>): Promise<void> => {
+export const cliSuppressExceptions = async (f: () => Promise<number>): Promise<void> => {
   try {
-    await f()
+    process.exitCode = await f()
```
veronicagg: is it [here](https://github.com/Azure/oav/blob/21d90189f5ba347d88784998bda244dde6b3f86c/lib/commands/validate-example.ts#L38) where validateExamples command exit code returns? if so will the exit code propagate out of the "handler" so it's returned as the cli exit code?

lib/commands/validate-spec.ts
sergey-shandar: It's a Boolean flag whish is `true` if the spec is valid and `false` otherwise. Why do we have it? I don't know.

lib/cliSuppressExceptions.ts
sergey-shandar: In this case, `f` is `validateExample`.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "oav",
+  "version": "0.16.1",
```
vladbarosan: duplicate

lib/validators/specValidator.ts
```diff
@@ -52,7 +52,7 @@ export interface ValidationResultScenarios {
 }
 
 export interface SpecValidationResult extends ModelValidation {
-  validityStatus: unknown
+  validityStatus: boolean
```
vladbarosan: can we leave it as unknown, ill update today in another PR with another breaking change

sergey-shandar: No, we need to change in this PR so we can use the flag.

lib/commands/validate-spec.ts
```diff
@@ -21,9 +22,11 @@ export async function handler(argv: yargs.Arguments): Promise<void> {
       pretty: argv.p
     }
     if (specPath.match(/.*composite.*/gi) !== null) {
-      await validate.validateCompositeSpec(specPath, vOptions)
+      const result = await validate.validateCompositeSpec(specPath, vOptions)
+      return iterator.some(result, v => !v.validityStatus) ? 1 : 0
```
veronicagg: could similar logic as my previous comment be applied here? is it required to check on validity status for each result?

lib/commands/validate-spec.ts
veronicagg: ok, is checking on results.length for this case not enough? do we need to check on validityStatus?

lib/commands/validate-example.ts
```diff
@@ -33,9 +34,11 @@ export async function handler(argv: yargs.Arguments): Promise<void> {
       pretty: argv.p
     }
     if (specPath.match(/.*composite.*/gi) !== null) {
-      await validate.validateExamplesInCompositeSpec(specPath, vOptions)
+      const result = await validate.validateExamplesInCompositeSpec(specPath, vOptions)
+      return iterator.flat(result).some(() => true) ? 1 : 0
```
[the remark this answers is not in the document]
sergey-shandar: No, it can't because it's an array of arrays. Some of them can be empty.

lib/commands/validate-spec.ts
sergey-shandar: No, if at least one is failed the job has to fail.

lib/validators/specValidator.ts
sergey-shandar: We need it as `Boolean` for this PR.

PR #411, opened 2019-04-05
Adding missing exports to index.ts

ericwill-msft:


index.ts
```diff
@@ -42,7 +42,11 @@ export {
   LiveResponse,
   LiveValidationIssue,
   LiveValidatorOptions,
-  ApiOperationIdentifier
+  ApiOperationIdentifier,
+  RequestValidationResult,
+  ResponseValidationResult,
+  ValidateOptions,
+  formatUrlToExpectedFormat
```
vladbarosan: I don't think we want to export "formatUrlToExpectedFormat" thats a function internal to the module

index.ts
```diff
@@ -42,7 +42,11 @@ export {
...
 } from "./lib/validators/liveValidator"
 export { SpecResolver } from "./lib/validators/specResolver"
```
vladbarosan: also lets update the `package.json` patch version otherwise publishing will fail

ericwill-msft: Updated to 0.15.4

index.ts
ericwill-msft: Removed

PR #427, opened 2019-05-30
Fix: Better error return when cannot find a reference

timotheeguerin:


lib/util/jsonUtils.ts
```diff
@@ -111,3 +111,11 @@ export async function parseJson(
 
   return swaggerObjectPromise
 }
+
+const getSpecContent = async (specPath: string) => {
+  try {
+    return await vfs.readFile(specPath)
+  } catch (error) {
+    throw new Error(`Failed to load a reference file ${specPath}. (${error})`)
```
vladbarosan: is this actually going to show anythign useful or we need to serialize it somehow / pick some message/code

timotheeguerin: Will show `(Error: StatusCode: "404", ResponseBody: "404: Not Found
.")` in that case. so the error message

PR #438, opened 2019-07-08
Update pathInPayLoad for MISSING_REQUIRED_PROPERTIES and OBJECT_ADDITONAL_PROPERTIES

lirenhe:
- Add additional check for  MISSING_REQUIRED_PROPERTIES and for OBJECT_ADDITONAL_PROPERTIES so pathInPayLoad should point to the property.
- Update unit test for LiveValidation for this change
- Fix the load problem of single file in unit tests.


lib/validators/liveValidator.ts
```diff
@@ -372,11 +372,20 @@ export class LiveValidator {
...
+        err.params.map((param: any) => err.path.concat(`/${param}`)).concat(err.similarPaths || []) : []
+    }
+    else{
+      pathsInPayload = err.path ? [err.path, ...(err.similarPaths || [])] : [];
+    }
+
     return {
       code: err.code || "INTERNAL_ERROR",
       message: err.message || "",
       jsonPathsInPayload: err.jsonPath ? [err.jsonPath, ...(err.similarJsonPaths || [])] : [],
```
ericwill-msft: I think we need to make sure jsonPathsInPayload is updated as well

lib/validators/liveValidator.ts
```diff
@@ -372,11 +372,20 @@ export class LiveValidator {
     }
   }
   private toLiveValidationIssue(err: { [index: string]: any; url: string }): LiveValidationIssue {
+    let pathsInPayload = [];
+    if(err.code === "OBJECT_MISSING_REQUIRED_PROPERTY" || err.code === 'OBJECT_ADDITIONAL_PROPERTIES'){
+      pathsInPayload = err.path ? 
+        err.params.map((param: any) => err.path.concat(`/${param}`)).concat(err.similarPaths || []) : []
```
ericwill-msft: Do we need to update similarPaths as well?

lib/validators/liveValidator.ts
```diff
@@ -372,11 +372,20 @@ export class LiveValidator {
     }
   }
   private toLiveValidationIssue(err: { [index: string]: any; url: string }): LiveValidationIssue {
```
vladbarosan: these errors can happen also in model validation, i would suggest is to apply this for that also

vladbarosan: we might be able to centralize that when the error is created in Sway

lirenhe: I checked the code and the path is built from the error and its inner error recursively.
So I don't think modifying path could be done easily in Sway without regression. It is much safer to handle it on the top level.

vladbarosan: yes, as we checked yesterday did you try centralizing it in `setPathProperties` properties. that way we we minimize the changes and they will apply to model validation/ semantic validation and jsonpath and jsonpointer

test/__snapshots__/liveValidatorTests.ts.snap
```diff
@@ -608,14 +608,14 @@ Object {
         "documentationUrl": "",
         "inner": undefined,
         "jsonPathsInPayload": Array [
-          "$.properties.groups[0]",
+          "$.properties.groups[0].displayName",
```
vladbarosan: actually looking here it seems we are missing examples for multiple similar paths.

can we add for example at https://github.com/Azure/oav/blob/master/test/liveValidation/payloads/objectAdditionalProperties_input.json#L46 another array entry ( can be a copy paste of the existing one ).
Then in the snapshots for that error we should get 2 paths.

vladbarosan:
looks mostly good, also @lirenhe can we bump the minor version ( not just the patch version) since this is a breaking change. Also please update the changelog.

regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -1506,19 +1506,6 @@ Array [
     "severity": 0,
     "source": "response",
   },
-  Object {
```
lirenhe: no sure why this part is removed from the snapshot. Any idea?

vladbarosan: i think it is because you are bringing an updated version of sway which contains a bug fix from Ray. So it should be fine.

vladbarosan:
@lirenhe  looks good, only thing is, please bump the minor version instead of the patch version as this is technically a breaking change.

Issue #386, opened 2019-02-05
OAV (sway) should report an error if `discriminator` is not required


sergey-shandar:

mozehgir:
Should be valid for all cases other than some edge cases like Cognitive Services who want to allow missing discrimantors.

PR #439, opened 2019-07-08
Report error if 'discriminator' is not required

raych1:
Issue #386 OAV report error if the property for 'discriminator' is not a
requried property.


lib/validators/suppressions.ts
```diff
@@ -20,8 +21,36 @@ export const getSuppressions = async (specPath: string): Promise<undefined | amd
...
 }
+
+export function existSuppression(
+  specPath: string,
+  suppression: amd.Suppression,
+  id: string
+): boolean {
+  if (suppression.directive !== undefined) {
+    const suppressionArray = getSuppressionArray(specPath, suppression.directive)
+    for (const s of suppressionArray) {
```
sergey-shandar:
```ts
return it.some(suppressionArray, s => s.suppress === id)
```

raych1: Updated.

PR #440, opened 2019-07-08
fix validation for nullable array types and fix a discrepancy in specResolver

### LINKED ISSUE #437: Setting x-nullable on array field
I'm trying to create an array field which could accept `null` (using `x-nullable: true`), but when I run the example validations I'm getting back errors (which don't occur when I replace the array type to a string for example).

I've created a short example that demonstrates the issue.

**Spec file:**
```json
{
    "swagger": "2.0",
    "schemes": [
        "https"
    ],
    "consumes": [
        "application/json"
    ],
    "produces": [
        "application/json"
    ],
    "paths": {
        "/x": {
            "patch": {
                "x-ms-examples": {
                    "Update": {
                        "$ref": "./examples/myExample.json"
                    }
                },
                "operationId": "_Update",
                "parameters": [{
                    "name": "data",
                    "in": "body",
                    "required": true,
                    "x-ms-parameter-location": "method",
                    "type": "object",
                    "schema": {
                        "type": "object",
                        "properties": {
                            "myObject": {
                                "$ref": "#/definitions/compositeProps"
                            }
                        }

                    }

                }],
                "responses": {
                    "200": {
                        "description": "Updated",
                        "schema": {
                            "$ref": "#/definitions/model"
                        }
                    }
                }
            }
        }
    },
    "definitions": {
        "model": {
            "type": "object",
            "properties": {
                "properties": {
                    "x-ms-client-flatten": true,
                    "$ref": "#/definitions/myProperties"
                }
            }

        },
        "myProperties": {
            "type": "object",
            "properties": {
                "myObject": {
                    "$ref": "#/definitions/compositeProps"
                }
            }
        },
        "compositeProps": {
            "type": "object",
            "properties": {
                "myArray": {
                    "type": "array",
                    "x-nullable": true,
                    "items": {
                        "type": "string"
                    }
                }
            },
            "required": [
                "myArray"
            ]
        }
    }
}
```

**Example file:**
```json
{
    "parameters": {
        "data": {
            "properties": {
                "myObject": {
                    "myArray": null
                }
            }
        }
    },
    "responses": {
        "200": {
            "body": {
                "properties": {
                    "myObject": {
                        "myArray": null
                    }
                }
            }
        }
    }
}
```

What am I missing? 🤔

amarzavery:
fixes #437


amarzavery:
@michaeljqzq - I realized that Ray assigned you to fix #437. However, I had already sent this PR by then. Please review this PR.

lib/validators/resolveNestedDefinitions.ts
```diff
@@ -76,7 +76,9 @@ export function resolveNestedDefinitions(spec: SwaggerObject): SwaggerObject {
   // a function to resolve SchemaObject (top-level and nested)
   const resolveSchemaObject = (schemaObject: SchemaObject): SchemaObject => {
     const result = propertySetMap<SchemaObject>(schemaObject, {
-      properties: properties => stringMapMap(properties, resolveNestedSchemaObject),
+      properties: skipIfUndefined(properties =>
```
raych1: Can you help me understand if this change caused by the change above of allowNullType set for top level? I don't see the relevance when try locally.

michaeljqzq: Per my understanding, this part is not quite related to above change

raych1:
@amarzavery , can you resolve the conflicts and merge this PR?

PR #442, opened 2019-07-11
@types/jsonpath and typescript 3.5

sergey-shandar:
`@types/jsonpath` should be a dependency. Otherwise we can't use some APIs from OAV:

```
node_modules/oav/dist/lib/util/validationError.d.ts(4,31): error TS7016: Could not find a declaration file for module 'jsonpath'. '/home/vsts/work/1/s/node_modules/jsonpath/index.js' implicitly has an 'any' type.
  Try `npm install @types/jsonpath` if it exists or add a new declaration (.d.ts) file containing `declare module 'jsonpath';`
```

https://dev.azure.com/azure-sdk/public/_build/results?buildId=64019&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774


package.json
```diff
@@ -46,23 +47,22 @@
...
     "jest": "^24.7.1",
     "jest-junit": "^6.3.0",
     "prettier": "^1.18.2",
     "should": "^13.2.3",
+    "ts-jest": "^24.0.2",
     "tslint": "^5.17.0",
     "tslint-config-prettier": "^1.18.0",
     "tslint-plugin-prettier": "^2.0.1",
-    "typescript": "~3.4.5"
+    "typescript": "^3.5.3"
```
raych1: Why do we change to caret ranges? Does typescript minor version change have breaking change?

sergey-shandar: We would like to automatically update to 3.6 when it's available. '^' is a kind of standard and default behavior. We introduce '~' only because we couldn't use `3.5` before but I fixed these issues.

Issue #423, opened 2019-04-30
OAV not reporting example position for INVALID_TYPE error in request


veronicagg:
https://github.com/Azure/azure-rest-api-specs/blob/f9f5d84cf22d860a02fb3188a58863c4241a89b6/specification/adhybridhealthservice/resource-manager/Microsoft.ADHybridHealthService/stable/2014-01-01/ADHybridHealthService.json

Output contains an INVALID_TYPE error in the REQUEST and points to the spec but not the corresponding position in the example 
Details:
Show inner error
Details: "Expected type object but found type array"
Spec location: specification/adhybridhealthservice/resource-manager/Microsoft.ADHybridHealthService/stable/2014-01-01/ADHybridHealthService.json#L6213

Operation ID - addsServices_update
Scenario = addsServices_update
Source= request
ErrorCode INVALID_TYPE

raych1:
@veronicagg here're the detail error message. It's clear to see the complained parameter name is 'service' though it doesn't have the line number of the example file. I would say won't fix. 
@amarzavery @sergey-shandar , do you have any comment on this issue? 
```
 code: 'REQUEST_VALIDATION_ERROR',
  id: 'OAV109',
  message:
   'Found errors in validating the request for x-ms-example "addsServices_update" in operation "addsServices_update".:\nFound errors in validating the request for x-ms-example "addsServices_update" in operation "addsServices_update".',
  innerErrors:
   [ { code: 'INVALID_REQUEST_PARAMETER',
       errors:
        [ { code: 'INVALID_TYPE',
            params: [ 'object', 'array' ],
            message: 'Expected type object but found type array',
            path: [],
            title:
             '{"path":["definitions","ServiceProperties"],"position":{"line":6213,"column":26},"url":"D:/git/github/azure-rest-api-specs/specification/adhybridhealthservice/resource-manager/Microsoft.ADHybridHealthService/stable/2014-01-01/ADHybridHealthService.json","directives":{}}',
            description: 'The service properties for a given service.' } ],
       in: 'body',
       message:
        'Invalid parameter (service): Value failed JSON Schema validation',
       name: 'service',
       schemaPath:
        '#/paths/~1providers~1Microsoft.ADHybridHealthService~1addsservices~1{serviceName}/patch/parameters/1' } ],
  level: '\u001b[31merror\u001b[39m' }
```

sergey-shandar:
[...]

PR #445, opened 2019-07-23
Report exmaple position for request errors

raych1:
Fixing [Issues#423](https://github.com/Azure/oav/issues/423) to report example position for INVALID_TYPE error in request.


lib/validators/specResolver.ts
```diff
@@ -453,10 +453,12 @@ export class SpecResolver {
         this.reportError,
         this.docsCache
       )
+      const result2: any = result
```
sergey-shandar: Please, don't use `any`. Use `unknown` if possible.

Could we have a better name instead of `result2` ?

lib/validators/modelValidator.ts
```diff
@@ -84,6 +88,22 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
...
+    try {
+      if (!this.exampleJsonMap.has(exampleFilePath)) {
+        const exampleJson = await jsonUtils.parseJson(
+          undefined,
+          exampleFilePath,
+          jsonParser.defaultErrorReport
+        )
+        this.exampleJsonMap.set(exampleFilePath, exampleJson)
+      }
+      // let position: sourcemap.FilePosition = sourcemap.getDescendantFilePosition(exampleJson, ["parameters","profileName"])
```
ruowan: Pls remove debug code here

lib/validators/modelValidator.ts
```diff
@@ -867,6 +898,29 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
...
+        requestValidationErrors !== null &&
+        requestValidationErrors.length > 0
+      ) {
+        if (exampleFilePath !== undefined) {
+          requestValidationErrors.forEach(error => {
+            const position = getDescendantFilePosition(
+              this.exampleJsonMap.get(exampleFilePath) as JsonRef,
+              error.path
+            )
+            error.title =
```
ruowan: This title format is difficult to read. Merge into one line may be better.

lib/validators/modelValidator.ts
```diff
@@ -867,6 +898,29 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
     if (!isValid) {
       operationResult.isValid = false
       operationResult.request.isValid = false
+
+      if (
+        requestValidationErrors !== undefined &&
```
ruowan: Can we use `!!requestValidationErrors` instead of complex judgement.

ruowan: For detail, see [link](https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript)

PR #454, opened 2019-09-18
Add prefix to log entry to disaply in github.

raych1:


lib/validate.ts
```diff
@@ -69,6 +69,12 @@ export async function getDocumentsFromCompositeSwagger(
   }
 }
 
+const vsoLogIssueWrapper = (issueType: string, message: string) => {
+  return issueType === "error" || issueType === "warning"
```
ruowan: Add return type.

PR #455, opened 2019-09-29
add unresolvable_reference into error codes

lirenhe:


raych1:
Please simplify `employee.json` or use a simple spec file for this `unresolvable_reference` test.

PR #457, opened 2019-10-14
Improve logging for livevalidator

raych1:
Added an optional logging callback to consume by livevalidator to dump
logs to external caller.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,9 +3,6 @@
 ## 10/14/2019 0.19.8
 
 - Add an optional logging callback to consume by livevalidator in order to dump log to external caller.
-
-## 10/12/2019 0.19.7
-
 - Update the yasway version to 1.8.5.
 - Add the new validation error code of SECRET_PROPERTY.
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 10/14/2019 0.19.8
+
+- Add an optional logging callback to consume by livevalidator in order to dump log to external caller.
+
+## 10/12/2019 0.19.7
+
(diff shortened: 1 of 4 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 10/14/2019 0.19.8
+
+- Add an optional logging callback to consume by livevalidator in order to dump log to external caller.
+
+## 10/12/2019 0.19.7
```
lirenhe: I didn't bump the version to 0.19.7 for those 2 change. 
Could you merge them together with 0.19.8

lib/validators/liveValidator.ts
```diff
@@ -158,6 +165,12 @@ export class LiveValidator {
     this.options = ops as LiveValidatorOptions
   }
 
+  private loggingCallback(message: string, level: number = winston.config.npm.levels.info): void {
```
lirenhe: Could you use string as level directly?  I don't think oav need to depend on winston

lib/validators/liveValidator.ts
```diff
@@ -656,7 +691,11 @@ export class LiveValidator {
         onlyFiles: true,
         unique: true
       })
-      log.debug(
+      log.info(
```
lirenhe: Could you merge log call and loggingCallback call into one private function?

ChangeLog.md
raych1: Updated.

lib/validators/liveValidator.ts
raych1: Updated.

lib/validators/liveValidator.ts
raych1: Resolved.

PR #458, opened 2019-10-22
Extends logging and metrics.

raych1:
Add corrleationId;
Log execution time for each operation;
Log splited request info for each call.

### DIFF
--- a/lib/validators/liveValidator.ts
+++ b/lib/validators/liveValidator.ts
@@ -213,7 +213,7 @@ export class LiveValidator {
     this.logging("Cache initialization complete.")
     const elapsedTime = Date.now() - startTime
     this.logging(
-      `DurationInMs:${elapsedTime}`,
+      `Cache initialization complete with DurationInMs:${elapsedTime}.`,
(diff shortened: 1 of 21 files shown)


lib/validators/liveValidator.ts
```diff
@@ -202,151 +211,11 @@ export class LiveValidator {
...
-    }
-
-    return new models.PotentialOperationsResult(
-      potentialOperations,
-      provider,
-      apiVersion,
-      liveValidationError
+    const elapsedTime = Date.now() - startTime
+    this.logging(
+      `DurationInMs:${elapsedTime}`,
```
lirenhe: Please provide a more meaningful description for this message.

lib/validators/liveValidator.ts
```diff
@@ -706,7 +821,14 @@ export class LiveValidator {
     })
 
     try {
+      const startTimeInnerValidator = Date.now()
```
lirenhe: Rename this param. I don't think this is a validator

PR #461, opened 2019-11-20
Resource provider and API-version should be case-insensitive for live validation

nickzhums:
After inspecting the code, I found the following.

**During Cache Initialization**
Both provider and api version will be converted to lower case when cache is being built in
`async getSwaggerInitializer(swaggerPath)`
`   let apiVersion = api.info.version.toLowerCase();`
`provider = provider.toLowerCase();`

**During validating**, provider is already being converted to lowercase, however api-version IS NOT. This causes problems if api-version is like "2018-01-01-STABLE / stable"
` apiVersion = queryObject["api-version"] as string
      providerNamespace = utils.getProvider(pathStr) || C.unknownResourceProvider
      resourceType = utils.getResourceType(pathStr, providerNamespace)

      // Provider would be provider found from the path or Microsoft.Unknown
      providerNamespace = providerNamespace || C.unknownResourceProvider
      if (providerNamespace === C.unknownResourceProvider) {
        apiVersion = C.unknownApiVersion
      }
      providerNamespace = providerNamespace.toLowerCase()
      queryStr = parsedUrl.query
      requestMethod = requestMethod.toLowerCase()`

Therefore changes need to be made for 

1. During validation, api-version needs to be converted to lower case
2. Add unit tests


test/invalidExamplesTests.ts
```diff
@@ -17,6 +17,6 @@ describe("Invalid examples", () => {
     if (result[0].details === undefined) {
       throw new Error("result[0].details === undefined")
     }
-    assert.notStrictEqual(result[0].details.url, undefined)
```
raych1: What about keep the origin asset too?

lib/validators/liveValidator.ts
```diff
@@ -667,6 +667,7 @@ export class LiveValidator {
         apiVersion = C.unknownApiVersion
       }
       providerNamespace = providerNamespace.toLowerCase()
+      apiVersion = apiVersion.toLowerCase()
       queryStr = parsedUrl.query
```
raych1: Can we make queryStr lowercase too for consistency?

test/invalidExamplesTests.ts
nickzhums: The original test was failing on this line, it's strange because I didn't modify any of related code and it's also failing on master branch. During debugging it seems like the data structure has changed and this test is no longer valid. Let me know what you think

lib/validators/liveValidator.ts
nickzhums: Sure

nickzhums: Sure, I'll make the change. It seems like in previous code block the function converts all keys to lowercase but not values

nickzhums: Updated

PR #466, opened 2019-12-05
Task #1032251 live validator swaggerPathPattern

raych1:
Change the swaggerPathPattern to array type from string.

### DIFF
--- a/lib/validators/liveValidator.ts
+++ b/lib/validators/liveValidator.ts
@@ -781,7 +781,7 @@ export class LiveValidator {
     return op as OperationWithApiVersion
   }
 
-  private async getMatchedPaths(jsonsPattern: string): Promise<string[]> {
+  private async getMatchedPaths(jsonsPattern: string | string[]): Promise<string[]> {
     const matchedPaths = await globby(jsonsPattern, {
       ignore: [
         "**/examples/**/*",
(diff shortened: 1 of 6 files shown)


lib/validators/liveValidator.ts
```diff
@@ -781,6 +781,26 @@ export class LiveValidator {
     return op as OperationWithApiVersion
   }
 
+  private async getMatchedPaths(jsonsPattern: string): Promise<string[]> {
+    const matchedPaths = await globby(jsonsPattern, {
```
nickzhums: is it possible to directly pass an array of strings to globby? if so maybe we can avoid concating the swagger paths

raych1: Good suggestion, changed it.

lib/validators/liveValidator.ts
```diff
@@ -36,7 +36,7 @@ export interface LiveValidatorOptions {
   }
   useRelativeSourceLocationUrl?: boolean
   directory: string
-  swaggerPathsPattern: string
+  swaggerPathsPattern: string[]
```
lirenhe: Can this be string | string[]?

lib/validators/liveValidator.ts
```diff
@@ -781,35 +781,48 @@ export class LiveValidator {
...
       )
       return this.options.swaggerPaths
     } else {
-      const allJsonsPattern = "/specification/**/*.json"
-      const jsonsPattern = path.join(
-        this.options.directory,
-        this.options.swaggerPathsPattern || allJsonsPattern
-      )
-      const swaggerPaths = await globby(jsonsPattern, {
-        ignore: [
```
lirenhe: Why is this removed?

lib/validators/liveValidator.ts
raych1: It can but it's not allowed to pass by chart in validation service.

PR #467, opened 2019-12-06
Exclude certain swaggers during cache initialization in OAV

nickzhums:
Per discussion about performance improvement of live validation, one area is to decrease the number of  Swagger specs to be loaded during cache initialization.
After testing with this branch, it looks like with this code change the cache build time can be reduced by ~ 5 mins, and memory can be reduced by 800mb ~ 1 g (All tested on AKS)


lib/util/constants.ts
```diff
@@ -124,3 +124,34 @@ export const unknownResourceProvider = "microsoft.unknown"
...
+    "**/imds/data-plane/**/*",
+    "**/iotcentral/data-plane/**/*",
+    "**/keyvault/data-plane/**/*",
+    "**/machinelearningservices/data-plane/**/*",
+    "**/monitor/data-plane/**/*",
+    "**/operationalinsights/data-plane/**/*",
+    "**/search/data-plane/**/*",
+    "**/servicefabric/data-plane/**/*",
+    "**/storage/data-plane/**/*",
+    "**/timeseriesinsights/data-plane/**/*"
```
nickzhums: by explicitly defining the paths, this ensures we don't exclude anything by accident, but also it might mean that people need to add path here as they go

lib/util/constants.ts
```diff
@@ -124,3 +124,34 @@ export const unknownResourceProvider = "microsoft.unknown"
...
+    "**/live/**/*",
+    "**/wire-format/**/*",
+    "**/azurestack/**/*",
+    "**/applicationinsights/data-plane/**/*",
+    "**/batch/data-plane/**/*",
+    "**/cognitiveservices/data-plane/**/*",
+    "**/containerregistry/data-plane/**/*",
+    "**/datalake-analytics/data-plane/**/*",
+    "**/datalake-store/data-plane/**/*",
+    "**/ventgrid/data-plane/**/*",
```
raych1: Typo of eventgrid?

test/liveValidatorTests.ts
```diff
@@ -18,6 +18,7 @@ describe("Live Validator", () => {
     it("should initialize with defaults", () => {
       const options = {
         swaggerPaths: [],
+        excludedSwaggerPathsPattern: Constants.DefaultConfig.ExcludedSwaggerPathsPattern,
```
raych1: Why do we need to add the default value assignment for each test?

lib/util/constants.ts
nickzhums: good catch

test/liveValidatorTests.ts
nickzhums: because the test is trying to verify the default options. Since we added the default excluded swagger patterns, we have to also add it in the test to match the behavior

test/liveValidatorTests.ts
```diff
@@ -226,6 +232,42 @@ describe("Live Validator", () => {
       await validator.initialize()
       assert.notStrictEqual(validator.cache["microsoft.batch"], undefined)
     })
+    it("should initialize and ignore certain swaggers by default", async () => {
```
raych1: Please update the description of this test.

test/liveValidatorTests.ts
```diff
@@ -226,6 +232,42 @@ describe("Live Validator", () => {
...
+      if (microsoftBatch === undefined) {
+        throw new Error("microsoftBatch === undefined")
+      }
+      // Should find two different api versions, one for data plane and one for management plane
+      assert.strictEqual(1, microsoftBatch["2019-08-01.10.0"].get.length)
+      assert.strictEqual(7, microsoftBatch["2017-01-01"].get.length)
+      assert.strictEqual(2, microsoftBatch["2017-01-01"].patch.length)
+      assert.strictEqual(4, microsoftBatch["2017-01-01"].post.length)
+    })
     it("should initialize for all swaggers", async () => {
```
raych1: As synced offline, pls add a test to verify the assigned excludedSwaggerPathsPattern value.

PR #472, opened 2019-12-30
oav support refresh cache per resource provider

ruowan:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.20.4",
+  "version": "0.20.4-beta.0",
```
raych1: Let's bump minor version for this change.

test/utilsTests.ts
```diff
@@ -43,4 +43,24 @@ describe("Utility functions", () => {
       assert.strictEqual(provider, "Microsoft.Authorization")
     })
   })
+
+  describe("Get provider by swagger file name", () => {
+    it("should return undefined", () => {
+      assert.equal(utils.getProviderBySwaggerFileName(""), undefined)
+    })
+    it("should return resource provider", () => {
+      assert.equal(
```
raych1: Can we change this test to negative case?

lib/validators/liveValidator.ts
```diff
@@ -825,7 +825,20 @@ export class LiveValidator {
     }
   }
 
-  private async getSwaggerInitializer(swaggerPath: string): Promise<void> {
+  public async refreshResourceProviderCache(resourceProvider: string): Promise<void> {
```
raych1: Suggest to pass in string array.

lib/validators/liveValidator.ts
```diff
@@ -825,7 +825,20 @@ export class LiveValidator {
...
+    const specsPaths = (await this.getSwaggerPaths()).filter(
```
raych1: Leverage options.swaggerPathsPattern which will be used to match the swagger to load.

lib/validators/liveValidator.ts
```diff
@@ -814,30 +814,42 @@ export class LiveValidator {
...
+        const patterns = pathsPatterns.map(item => path.join(this.options.directory, item))
+        return this.getMatchedPaths(patterns)
       }
     }
   }
 
-  private async getSwaggerInitializer(swaggerPath: string): Promise<void> {
+  public async refreshResourceProviderCache(resourceProviders: string[]): Promise<void> {
+    const pathsPatterns = resourceProviders.map(
+      it => `/specification/**/resource-manager/${it}/**/*.json`
```
lirenhe: Could you use a better name to replace 'it'?

lib/util/utils.ts
```diff
@@ -210,6 +210,20 @@ export function setObject(doc: {}, ptr: string, value: unknown, overwrite = true
...
+ * @example input: apimanagement/resource-manager/Microsoft.ApiManagement/preview/2018-01-01/apimwriteonly.json.
+ * return: Microsoft.ApiManagement
+ */
+export function getProviderBySwaggerFileName(filename: string): string | undefined {
+  const pattern = new RegExp("resource-manager/(?<resourceProvider>.*?)/")
+  const matchRes = filename.match(pattern)
+  if (matchRes === null) {
+    return undefined
+  }
+  return matchRes.groups === undefined ? undefined : matchRes.groups.resourceProvider
```
lirenhe: Could you refine this line?

ruowan: Use ts3.7 optional chaining.

lib/validators/liveValidator.ts
ruowan: updated

lib/util/utils.ts
ruowan: We can use optional chaining to instead it after upgrade typescript to 3.7.4 .

lib/validators/liveValidator.ts
```diff
@@ -835,15 +835,33 @@ export class LiveValidator {
...
   public async refreshResourceProviderCache(resourceProviders: string[]): Promise<void> {
     const pathsPatterns = resourceProviders.map(
-      it => `/specification/**/resource-manager/${it}/**/*.json`
+      resourceProvider => `/specification/**/resource-manager/${resourceProvider}/**/*.json`
     )
-    const specsPaths = await this.getSwaggerPaths(pathsPatterns)
-    const promiseFactories = specsPaths.map(swaggerPath => {
-      return this.getSwaggerInitializer(swaggerPath, true)
+    resourceProviders.forEach(resourceProvider => {
+      delete this.cache[resourceProvider.toLowerCase()]
```
raych1: The cache item cannot be deleted and has to be updated in place.

ruowan: Update equals delete and reload.

raych1:
Close as this is out of date.

PR #473, opened 2020-01-03
Update the check for MultipleOperationsFound

lirenhe:
Improve the error handling and update the test case


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.20.4",
+  "version": "0.20.5-rc",
```
raych1: Use beta for testing purpose instead of rc.

test/liveValidatorTests.ts
```diff
@@ -673,6 +673,29 @@ describe("Live validator snapshot validation", () => {
...
+      }
+    }
+    const liveValidator = new LiveValidator(options)
+    await liveValidator.initialize()
+
+    const payload = require(`${__dirname}/liveValidation/payloads/multiplePperationFound_input`)
+    const result = liveValidator.validateLiveRequestResponse(payload)
+    expect(
+      result.responseValidationResult.runtimeException &&
+        result.responseValidationResult.runtimeException.code === "MULTIPLE_OPERATIONS_FOUND"
```
raych1: Can we verify the error message contain the specPath?

test/liveValidation/swaggers/specification/mediaservices/resource-manager/Microsoft.Media/2019-05-01-preview/Common.json
```diff
@@ -0,0 +1,80 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Azure Media Services",
+    "version": "2018-07-01"
```
raych1: Should this version be same as folder name?

lirenhe: No, it is to create an issue of multiple operations in the same API version and this is actually a real case.

Issue #468, opened 2019-12-23
Discriminator property value is resolved to parent model name


raych1:
This issue is detected by [PR#7907](https://github.com/Azure/azure-rest-api-specs/pull/7907).
Issue description:
When resolve discriminator in polymorphic definitions, it would set the discriminator property value of parent model definition(baseclass) equal to the parent model definition name instead of the defined enum values. 
e.g.  below `criterionType` property is resolved as `enum['MultiMetricCriteria']`
```code
"MultiMetricCriteria": {
      "type": "object",
      "additionalProperties": {
        "type": "object"
      },
      "discriminator": "criterionType",
      "properties": {
        "criterionType": {
          "enum": [
            "StaticThresholdCriterion",
            "DynamicThresholdCriterion"
          ],
          "x-ms-enum": {
            "name": "criterionType",
            "modelAsString": true
          },
          "description": "Specifies the type of threshold criteria"
        },
      },
      "required": [
        "criterionType"
      ]
    }
```
The related code to process discriminator is around [here](https://github.com/Azure/oav/blob/7c40ac85724f9108f541626123ebd9e63c2437c1/lib/validators/specResolver.ts#L738)

raych1:
This issue was fixed in the linked PR.

PR #475, opened 2020-02-05
Base model discriminator value fix.

raych1:
[Issue#468](https://github.com/Azure/oav/issues/468)
Previously for polymorphic case, the discriminator value was set to
model name in case of base model. It should also keep original enum
values which is used in model validation. So this fix is to keep
original enum values then add base model as first element in the enum
values list.
The reason to put base model name as first element is the first element
is used to match the discriminator value of polymorphic models and for
base model case the discriminator should be base model name.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 02/05/2020 0.20.7
+- Add base model name as a value to discriminator enum value list.[Issue#468](https://github.com/Azure/oav/issues/468)
+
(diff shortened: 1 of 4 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 02/05/2020 0.20.7
+- Add base model name as a value to discriminator enum value list.[Issue#468](https://github.com/Azure/oav/issues/468)
```
lirenhe: Please provide test case for this fix

PR #477, opened 2020-02-13
The path matching for live validation should be case insensitive

lirenhe:


ChangeLog.md
```diff
@@ -1,25 +1,36 @@
 # Changelog
 
+## 02/13/2020 0.20.8
+
+- The path matching for Live validation should be case-insensitive.
```
raych1: Can you specify the change details? It's not whole path matching change.

PR #478, opened 2020-02-13
Fall back to child resouce match for liveValidator

raych1:
Fall back to match operations in cache to child resouces if having both
of parent and child resouces in request url.

### DIFF
--- a/lib/util/utils.ts
+++ b/lib/util/utils.ts
@@ -276,6 +276,28 @@ export function getResourceType(pathStr: string, provider?: string): string {
 }
 
 /**
+ * Gets last child resouce url to match.
+ * @param {string} requestUrl The request url.
+ *                 Example "/subscriptions/randomSub/resourceGroups/randomRG/providers/providers/Microsoft.Storage/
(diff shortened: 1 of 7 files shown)


lib/util/utils.ts
```diff
@@ -275,6 +275,28 @@ export function getResourceType(pathStr: string, provider?: string): string {
...
+ *                 privatecontainer/providers/Microsoft.Authorization/roleAssignments/3fa73e4b-d60d-43b2-a248-fb776fd0bf60"
+ *                 will return "roleAssignments".
+ *
+ * @returns {string} last child resource url.
+ */
+export function getLastResourceUrlToMatch(requestUrl: string): string {
+  let index = requestUrl.lastIndexOf("/providers")
+  if (index > 0) {
+    const originUrlWithoutLastChildResource = requestUrl.substring(0, index)
+    index = originUrlWithoutLastChildResource.lastIndexOf("/")
```
lirenhe: why do we need the item behind the /providers

raych1: All the url after provider needs to be used to match operation.

PR #479, opened 2020-03-01
Upgrade virtual-fs pkg version.

raych1:


lib/validate.ts
```diff
@@ -70,9 +70,11 @@ export async function getDocumentsFromCompositeSwagger(
 }
 
 const vsoLogIssueWrapper = (issueType: string, message: string) => {
-  return issueType === "error" || issueType === "warning"
-    ? `##vso[task.logissue type=${issueType}]${message}`
-    : `##vso[task.logissue type=error]${message}`
+  if (issueType === "error" || issueType === "warning") {
+    return `##vso[task.logissue type=${issueType}]${message}`
+  } else {
+    return `##vso[task.logdetail]${message}`
```
lirenhe: why is this change?

PR #481, opened 2020-03-10
Fixed typeerror when search in unknownapiversion.

raych1:
added check when search in unknownrp->unknownapiversion;
added unit test.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 03/02/2020 0.20.12
 
-- Fixed typeerror when search in unknownrp-unknownapiversion cached operations
+- Fixed typeerror in live validator when search operation in case of unknownapiversion
 
 ## 03/02/2020 0.20.11
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 03/02/2020 0.20.12
+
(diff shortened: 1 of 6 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 03/02/2020 0.20.12
+
+- Fixed typeerror when search in unknownrp-unknownapiversion cached operations
```
lirenhe: could you refactor this sentence? I don't quite get the meaning.

lib/validators/liveValidator.ts
```diff
@@ -742,16 +742,18 @@ export class LiveValidator {
...
-          const pathObject = operation.pathObject
-          let pathTemplate = pathObject.path
-          if (pathTemplate && pathTemplate.includes("?")) {
-            pathTemplate = pathTemplate.slice(0, pathTemplate.indexOf("?"))
-            pathObject.path = pathTemplate
-          }
-          const pathMatch = pathObject.regexp.exec(requestInfo.pathStr)
-          return pathMatch !== null
-        })
+        if (operations && !operations.length) {
```
lirenhe: Could you use (operation.length>0) directly?

raych1: It has same result as !operations.length. Here only needs to check it's not 0 for the length. Keep this usage as other places too.

PR #482, opened 2020-03-10
fix  issue caused by host path parameter value with url

jianyexi:


lib/validators/modelValidator.ts
```diff
@@ -784,9 +784,13 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
...
           }
-          // replacing forward slashes with empty string because this messes up Sways regex
+
+          // replacing forward slashes and scheme with empty string because this messes up Sways regex
           // validation of path segment.
-          parameterValue = parameterValue.replace(/\//gi, "")
+          if (parameterValue.startsWith(scheme + "://")) {
+            parameterValue = parameterValue.slice(scheme.length + 3)
+          }
+          parameterValue = (parameterValue as string).replace(/^\//gi, "")
```
raych1: As synced offline to normalize the invalid characters.

lib/validators/modelValidator.ts
```diff
@@ -784,9 +784,14 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
...
           }
-          // replacing forward slashes with empty string because this messes up Sways regex
+
+          // replacing characters that may cause validator failed  with empty string because this messes up Sways regex
           // validation of path segment.
-          parameterValue = parameterValue.replace(/\//gi, "")
+          if (!utils.isUrlEncoded(parameterValue as string)) {
+            // TODO: we can get the scheme from parameterValue if the useSchemePrefix is setting false in the x-ms-parameterized-host,
+            // then check if it can match to the swagger scheme.
+            parameterValue = parameterValue.replace(/[^0-9a-zA-Z._]/gi, "")
```
raych1: Can we just replace `:` and `/` since there're other valid characters? such as `()&?` .

jianyexi: Due to the result of  test , these characters are also invalid : ()&?
Since I only remain  the characters that are valid in urlencode except *+- , can ensure the value is valid,other

raych1: Let's use whitelist as is now since the value is not validated furthermore.

ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 03/13/2020 0.20.13
+
+- Normanormalize the path parameter values by using whilelist (0-9a-zA-Z._) to fix the issue of validate-example failed
```
raych1: Can you fix typo `Normalize`?

jianyexi: done

raych1:
Seems the check was hung. Please make sure to merge with squash the commits.

raych1:
@jianyexi , I worked around the pipeline issue. Pls fix the snapshot issue.

jianyexi:
> @jianyexi , I worked around the pipeline issue. Pls fix the snapshot issue.

Done

PR #489, opened 2020-03-12
Fix invalid internal reference can not report error

jianyexi:


lib/validators/specResolver.ts
```diff
@@ -1027,4 +1033,25 @@ export class SpecResolver {
     }
     return result
   }
+
+  /**
+   * Check if exist undefined within-document reference
+   */
+  private checkWithInReferences() {
```
raych1: Can we change the name to `VerifyInternalReference`?

lib/validators/specResolver.ts
```diff
@@ -218,6 +220,10 @@ export class SpecResolver {
         this.resolveNullableTypes()
       }
     } catch (err) {
+      // to avoid double wrap the exception
+      if (typeof err === "object" && err.id && err.message) {
```
raych1: It needs to wrap the error message as `internal error`.

lib/validators/specResolver.ts
```diff
@@ -1027,4 +1033,25 @@ export class SpecResolver {
...
+  /**
+   * Check if exist undefined within-document reference
+   */
+  private checkWithInReferences() {
+    const errsDetail: any[] = []
+    const unresolvedRefs = jsonUtils.findUndefinedWithinDocRefs(this.specInJson)
+    unresolvedRefs.forEach((pathStr, ref) => {
+      const err: any = {}
+      err.path = (pathStr as string[]).join(".")
+      err.message = `JSON Pointer points to missing location:${ref}`
```
raych1: How about change the message to `The reference cannot be resolved`?

test/liveValidation/swaggers/specification/rpsaas/resource-manager/Microsoft.Contoso/stable/2019-01-01/employee.json
```diff
@@ -192,7 +200,7 @@
           "200": {
             "description": "Operation details.",
             "schema": {
-              "$ref": "#/definitions/OperationsList"
+              "$ref": "#/definitions/NestedResource"
```
raych1: Why to change this? Is this used in other tests?

test/liveValidation/swaggers/specification/batch/data-plane/Microsoft.Batch/stable/2019-08-01.10.0/BatchService.json
```diff
@@ -54,7 +54,7 @@
   },
   "paths": {
     "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Batch/batchAccounts/{accountName}/applications": {
-      "get": 
+      "get": {
```
raych1: Do you fix JSON format?

test/liveValidatorTests.ts
```diff
@@ -735,7 +735,7 @@ describe("Live validator snapshot validation", () => {
     )
   })
 
-  test(`should return expected error for unresolvable reference`, async () => {
+  test.skip(`should return expected error for unresolvable reference`, async () => {
```
raych1: Why the test is skipped?

test/semanticValidatorTests.ts
```diff
@@ -48,4 +48,10 @@ describe("Semantic validation", () => {
     const result = await validate.validateSpec(specPath, undefined)
     assert(result.validityStatus === true)
   })
+
+  it("should fail when validating a swagger with invalid reference", async () => {
```
raych1: Pls rename to `...invalid internal reference`.

test/liveValidation/swaggers/specification/rpsaas/resource-manager/Microsoft.Contoso/stable/2019-01-01/employee.json
jianyexi: Since I add internal reference check before operation validate and will throw an exception , this json file need to fix the internal error, or there are 5 test cases would not passed

lib/validators/specResolver.ts
jianyexi: Because sometimes  the error message  have already been wrapped as internal err

test/liveValidatorTests.ts
jianyexi: this test should skip because we have already detected the internal error before the sway do it

lib/validators/specResolver.ts
jianyexi: this is detail message , there is a  summary description  the same as you mentioned

lib/validators/specResolver.ts
jianyexi: fixed it

lib/validators/specResolver.ts
raych1: I know, but does all the error object here is `internal` err? If not, we still need to wrap them.

jianyexi: if the exception have already specify the id and message , It must be a exception that has been wrapped
so here we don't need to wrap it again

PR #502, opened 2020-04-10
fix model validation failure when path parameter value containing sep…

jianyexi:
…ecial charaters


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oav",
-  "version": "0.21.3",
+  "version": "0.21.4",
```
raych1, line 3: Can you update 'ms-rest' version in this file?

lib/validators/modelValidator.ts
```diff
@@ -790,10 +792,9 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
           // validation of path segment.
           parameterValue = parameterValue.replace(/\//gi, "")
 
-          // replacing scheme that may cause validator failed when x-ms-parameterized-host enbaled
-          if (hostTemplate) {
-            // TODO: we can get the scheme from parameterValue if the useSchemePrefix is setting false in the x-ms-parameterized-host,
-            // then check if it can match to the swagger scheme.
+          // replacing scheme that may cause validator failed when x-ms-parameterized-host enbaled & useSchemePrefix enabled
+          // because if useSchemePrefix enabled ,the parameter value in x-ms-parameterized-host should not has the scheme (http://|https://)
+          if (useSchemePrefix) {
```
raych1, line 797: useSchemePrefix is true by default if it's not set. 
https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md#x-ms-parameterized-host

Can you also add test case when useSchemePrefix = false?

jianyexi: sure, that's important

jianyexi: already has the test case in  describe : "for parameters in x-ms-parameterized-host"

test/modelValidatorTests.ts
```diff
@@ -264,6 +264,22 @@ describe("Model Validation", () => {
       assert(result.length === 0, `swagger "${specPath2}" contains model validation errors.`)
       // console.log(result)
     })
+
+    it("should pass for parameters include pattern", async () => {
```
raych1, line 268: can you add one more test for useSchemePrefix:false?

regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -13287,224 +13135,12 @@ exports[`validateExamples should not regress for file '/home/vsts/work/1/s/regre
...
-    },
-    "operationId": "Alerts_UpdateResourceGroupLevelAlertState",
-    "responseCode": "ALL",
-    "scenario": "Update security alert state on a resource group from a security data location",
-    "severity": 0,
-    "source": "request",
-  },
-]
-`;
+exports[`validateExamples should not regress for file '/home/vsts/work/1/s/regression/azure-rest-api-specs/specification/security/resource-manager/Microsoft.Security/stable/2019-01-01/alerts.json': returned results 1`] = `Array []`;
```
raych1, line 13138: Do you know why this new one is added?

PR #504, opened 2020-04-20
Output the exception information when pretty switch is enabled

raych1:
Pretty switch is used when call OAV from pipeline, output the log when exception happens.

### DIFF
--- a/lib/validate.ts
+++ b/lib/validate.ts
@@ -221,7 +221,28 @@ export async function validateExamples(
       }
       return errors
     } catch (e) {
-      log.error(e)
+      if (o.pretty) {
+        if (process.env["Agent.Id"]) {
+          console.log(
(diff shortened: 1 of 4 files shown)


lib/validate.ts
```diff
@@ -221,7 +221,28 @@ export async function validateExamples(
...
+            )
+          )
+          /* tslint:disable-next-line:no-console no-string-literal */
+          console.error(vsoLogIssueWrapper("error", e))
+        } else {
+          console.error(
+              `Validating "examples" and "x-ms-examples" in  ${specPath}:\n`
+          )
+          /* tslint:disable-next-line:no-console no-string-literal */
+          console.error("\x1b[31m", "error", ":", "\x1b[0m")
```
lirenhe, line 239: why using those unicodes  \x1b[31m?

PR #508, opened 2020-04-22
check out error code of snap files

Ligengxin96:
@raych1 I didn't see where can i add a reviewer when I commit a PR


regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -1107,6 +1107,158 @@ Array [
     "severity": 0,
     "source": "response",
   },
+  Object {
```
raych1, line 1110: This file doesn't need to update and can you remove it from this PR?

Ligengxin96: sorry, my fault, now this PR jsut modify the validateExamplesRegression_nonLatest.test.ts.snap

PR #526, opened 2020-08-17
Remove sensitive value is the live validation result of x-ms-secret (…

lirenhe:
…#525)

* Remove sensitive value is the live validation result of x-ms-secret

* Undo unnecessary file

Co-authored-by: FAREAST\renhel <renhel@microsoft.com>


raych1:
Use merge commit instead of squash merge.

PR #530, opened 2020-09-02
Update function getErrorsFromSemanticValidation to extract inner error correctly

Descatles:


raych1 (approved):

raych1:
@Descatles , can you also check the old pipeline result which will also call this function to parse validation result?

PR #543, opened 2020-10-15
Ignore required rule check in response.

raych1:
Only in case of both required and x-ms-secret:true are annotated.


lirenhe (approved):

raych1:
CI checks are not triggered.

PR #544, opened 2020-10-16
Fixed two bugs in error handling

raych1:
Output model detail in log when resolve it with failure,
report error when the example reference has issues.

### DIFF
--- a/lib/validators/modelValidator.ts
+++ b/lib/validators/modelValidator.ts
@@ -376,7 +376,8 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
       for (const [scenario, xmsExampleFunc] of sm.entries<any>(xmsExamples)) {
         if (typeof xmsExampleFunc !== "function") {
           const error = {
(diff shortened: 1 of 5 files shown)


lib/validators/modelValidator.ts
```diff
@@ -374,13 +374,21 @@ export class ModelValidator extends SpecValidator<SpecValidationResult> {
     const exampleFileMap = new Map<string, string>()
     if (xmsExamples) {
       for (const [scenario, xmsExampleFunc] of sm.entries<any>(xmsExamples)) {
-        const xmsExample = xmsExampleFunc()
-        resultScenarios[scenario] = {
-          requestValidation: this.validateRequest(operation, xmsExample.parameters),
-          responseValidation: this.validateXmsExampleResponses(operation, xmsExample.responses)
+        if (typeof xmsExampleFunc !== "function") {
```
jianyexi, line 377: Could you add a test case for it ?

jianyexi (approved):

lib/validators/modelValidator.ts
raych1: Added test.

PR #546, opened 2020-11-25
Support loading swagger examples

PhoenixHe-msft:


lib/swagger/jsonLoader.ts
```diff
@@ -106,11 +115,16 @@ export class JsonLoader implements Loader<Json> {
...
+    return this.mockNameMap[mockName];
+  }
+
   private async resolveRef(
     object: Json,
     pathArr: string[],
     rootObject: Json,
-    relativeFilePath: string
+    relativeFilePath: string,
+    skipResolveChildRef: boolean
```
raych1, line 127: why don't make this as optional?

raych1 (commented):
Please merge the change from develop.

lib/swagger/jsonLoader.ts
PhoenixHe-msft: It's a private function and I want to make it explicit to avoid bugs.

PhoenixHe-msft:
@raych1 Can we merge this?

raych1 (approved):

PR #548, opened 2020-12-18
Tih/always transform response code

PhoenixHe-msft:


raych1:
we don't want to merge the AJV related code to master.

PR #550, opened 2020-12-28
Example generation improvement

jianyexi:


lib/commands/generate-examples.ts
```diff
@@ -24,6 +24,16 @@ export const builder: yargs.CommandBuilder = {
     describe: "the directory path contains payload.",
     string: true,
   },
+  c: {
+    alias: "config",
+    describe: "config path.",
```
raych1, line 29: How about describe it as 'Readme config path'?

lib/generator/mocker.ts
```diff
@@ -12,7 +14,7 @@ export default class Mocker {
       case "array":
         return this.generateArray(paramSpec, paramName, arrItem);
       default:
-        console.log(`Unknown type ${paramSpec.type}.`);
+        log.info(`unknown type ${paramSpec.type}.`);
```
raych1, line 17: Are you going to update all other logging place to use `log.`?

lib/generator/swaggerMocker.ts
```diff
@@ -128,21 +154,15 @@ export default class SwaggerMocker {
...
-
-  private getSpecByDiscriminatorValue(value: string) {
-    const spec = this.getSpecDefinition(value);
-    if (spec) {
-      return spec;
-    }
-    for (const definition of Object.values(this.spec.definitions)) {
-      if ((definition as any)["x-ms-discriminator-value"] === value) {
-        return definition;
+  private getCache(schema:any) {
```
raych1, line 157: Do we need to pass in cacheType when try to get the value from cache?

lib/generator/swaggerMocker.ts
```diff
@@ -311,10 +366,24 @@ export default class SwaggerMocker {
         ...properties,
         ...this.getProperties(this.getDefSpec(item, visited), visited),
       };
+      this.removeFromSet(item,visited)
```
raych1, line 369: Can you add comment of reason to add this remove operation?

lib/commands/generate-examples.ts
jianyexi: Okay

lib/generator/swaggerMocker.ts
raych1: Let's rename payloadCache as exampleCache.

lib/generator/swaggerMocker.ts
raych1: To resolve a bug.

lib/generator/mocker.ts
jianyexi: Yes, already replaced almost all, left some to help understand the tool's running progress

lib/generator/exampleGenerator.ts
```diff
@@ -77,61 +113,137 @@ export default class Generator {
...
+    log.info(example);
     const unifiedExample = this.unifyCommonProperty(example);
-
     const newSpec = util.referenceExmInSpec(
       this.specFilePath,
       specItem.path,
       specItem.methodName,
-      `${operationId}_Base`
+      `${operationId}_${rule.exampleNamePostfix}_Base`
     );
```
raych1, line 196: Let's call the generated exmaple name as `{operationId}_{rule.postfix}_Gen`.

lib/generator/exampleGenerator.ts
```diff
@@ -77,61 +113,137 @@ export default class Generator {
...
+      specItem = this.getSpecItem(this.spec, operationId);
+      if (!specItem) {
+        console.error(`no specItem for the operation id ${operationId}`);
+        return [];
+      }
+    }
+    const ruleSet: RuleSet = []
+    ruleSet.push({
+      exampleNamePostfix: "Maximum",
+      ruleName: "Maximum"
```
raych1, line 234: Can you change it to `MaximumSet` for better understanding.

lib/generator/exampleGenerator.ts
```diff
@@ -77,61 +113,137 @@ export default class Generator {
...
+    });
+    ruleSet.push({
+      exampleNamePostfix: "Minimum",
+      ruleName: "Minimum"
```
raych1, line 238: Can you change it to `MinimumSet` for better understanding?

lib/generator/exampleRule.ts
```diff
@@ -1,7 +1,7 @@
 import { CacheItem } from "./exampleCache"
 export interface ExampleRule  {
```
raych1, line 2: Why not use type and make exampleNamePostfix as optional field?

jianyexi: The two filed are required,  using interface is considering that the rule may be extended in the future

raych1:
Can you update the existed examples which has `_Base` as postfix in the name?

raych1:
Can you fix check failures?

raych1 (approved):

lib/generator/exampleGenerator.ts
jianyexi: done

PR #563, opened 2021-01-20
Bugfix should copy required array for allOf transformer

PhoenixHe-msft:


ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Change Log - oav
 
+## 2.2.0
```
raych1, line 3: 2.2.1

lib/swaggerValidator/ajvDiscriminatorMap.ts
```diff
@@ -52,6 +52,8 @@ export const ajvEnableDiscriminatorMap = (ajv: Ajv, loader: JsonLoader) => {
               schemaPath: "/discriminator",
               schema: schemas,
               parentSchema,
+              _realSchema: schemas,
```
raych1, line 55: Please add some comment about the issue to fix and the how is it comes from and how is used.

raych1 (approved):

PR #564, opened 2021-01-22
fix undefined reference

jianyexi:


raych1:
Can you bump the version and add change log?

raych1 (approved):

PR #566, opened 2021-01-22
Support tenent operation with scope

PhoenixHe-TSP:


lib/transform/pathRegexTransformer.ts
```diff
@@ -62,7 +62,7 @@ const buildPathRegex = (
   params.forEach((v, i) => {
     if (path.startsWith(`/{${v}}`) && pathParams.get(v)) {
       // We allow first param in path as multi path param
-      path = path.replace("{" + v + "}", "(.*)");
+      path = path.replace("/{" + v + "}", "(.*)");
```
raych1, line 65: How to ensure it matches to the path of `/providers/xxx` firstly if it's defined for tenant scope?

raych1:
Please also add a unit test.

PR #579, opened 2021-02-25
Add new rule of missing_resource_id and isArmCall option.

raych1:
1. new rule of MISSING_RESOURCE_ID: all the GET/PUT resource call should have 'id' returned in response;
2. add an option isArmCall in liveValidator to differentiate the caller in order to enable different rulesets;
3. fix an issue in response header sourcemap.


lib/liveValidation/operationValidator.ts
```diff
@@ -128,7 +128,12 @@ export const validateSwaggerLiveResponse = async (
     validateContentType(operation.produces!, headers, false, result);
   }
 
-  const ctx = { isResponse: true, includeErrors, statusCode };
+  const ctx = {
+    isResponse: true,
+    includeErrors,
+    statusCode,
+    isGetPutHttpMethods: operation._method === "get" || operation._method === "put",
```
PhoenixHe-msft, line 135: How about rename to "httpMethod" to enable more future scenario

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -179,7 +183,7 @@ export const ajvErrorToSchemaValidateIssue = (
     message: errInfo.message,
     params: errInfo.params,
     jsonPathsInPayload: [dataPath],
-    schemaPath: err.schemaPath,
+    schemaPath: errInfo.code === "MISSING_RESOURCE_ID" ? "" : err.schemaPath,
```
PhoenixHe-msft, line 186: Why do we need to hard code this logic?

lib/transform/allOfTransformer.ts
```diff
@@ -18,6 +18,9 @@ const transformAllOfSchema = (schema: Schema, baseSchemas: Set<Schema>, jsonLoad
     transformAllOfSchema(sch, baseSchemas, jsonLoader);
 
     const { properties, required, additionalProperties: aProperties } = sch;
+    if (sch["x-ms-azure-resource"] === true) {
+      schema["x-ms-azure-resource"] = true;
```
PhoenixHe-msft, line 22: What if schema["x-ms-azure-resoutce"] is false? Do we intend to overwrite it?

PhoenixHe-msft: Consider use the variable xmsAzureResource to replace the string

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -50,6 +50,10 @@ export class AjvSchemaValidator implements SchemaValidator {
       ...options,
     });
     ajvEnableAll(this.ajv, loader);
+
+    if (options && (options as any).isArmCall === true) {
```
PhoenixHe-msft, line 54: Why options as any?

PhoenixHe-msft: Consider add interface SchemaValidatorOptions with isArmCall, then LiveValidatorOptions would inherit SchemaValidatorOptions

lib/liveValidation/operationValidator.ts
raych1: Refactored.

lib/swaggerValidator/ajvSchemaValidator.ts
raych1: Doesn't want to set it as '#/x-ms-azureresource' and leave it as ''.

lib/transform/allOfTransformer.ts
raych1: Override it only when undefined.

lib/swaggerValidator/ajvSchemaValidator.ts
raych1: LiveValidatorLoaderOptions cannot transform to ajv.Options.

PhoenixHe-msft: You need another option not for ajv.Options

PhoenixHe-msft (approved):

PR #587, opened 2021-03-12
Fixed mocked value of LRO headers

raych1:


lib/generator/swaggerMocker.ts
```diff
@@ -92,7 +92,7 @@ export default class SwaggerMocker {
       return;
     }
     return {
-      [headerAttr]: "LocationURl",
+      [headerAttr]: "https://foo.com/operationstatus",
```
jianyexi, line 95: Why change to this ?

raych1: autorest validates the value of these two headers which have to be a URL.

jianyexi (approved):

PR #588, opened 2021-03-15
User/raychen/lro

raych1:
Add new rules of 'LRO_RESPONSE_HEADER' and 'LRO_RESPONSE_CODE'.
Add option of 'isArmCall' to differentiate rulesets for Arm and RpaaS in validation apis.

See [document](https://github.com/Azure/azure-rest-api-specs/pull/13416) of these two rules for details.

### DIFF
--- a/lib/liveValidation/operationValidator.ts
+++ b/lib/liveValidation/operationValidator.ts
@@ -267,26 +267,48 @@ const validateLroOperation = (
   result: LiveValidationIssue[]
 ) => {
   if (operation["x-ms-long-running-operation"] === true) {
-    if (
-      operation._method === "delete" ||
-      operation._method === "patch" ||
-      operation._method === "post"
-    ) {
-      if (statusCode !== "202") {
-        result.push(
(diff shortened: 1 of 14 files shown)


lib/liveValidation/operationValidator.ts
```diff
@@ -256,6 +260,58 @@ const schemaValidateIssueToLiveValidationIssue = (
...
 
+const validateLroOperation = (
+  operation: Operation,
+  statusCode: string,
+  headers: StringMap<string>,
+  result: LiveValidationIssue[]
+) => {
+  if (operation["x-ms-long-running-operation"] === true) {
+    if (operation._method === "patch" || operation._method === "post") {
+      if (statusCode !== "202" && statusCode !== "201") {
```
jianyexi, line 271: Is the statusCode here always 2xx ?

raych1: Good catch, excluded failure cases.

jianyexi (approved):

PR #605, opened 2021-04-01
Support extension resource validation in live validator

raych1:
The path of extension resource starts with parameter, e.g. `/{resourceUri}/providers/Microsoft.Sc/linkers/{linkerName}`


test/liveValidatorTests.ts
```diff
@@ -558,6 +558,24 @@ describe("Live Validator", () => {
         // console.dir(validationResult, { depth: null, colors: true })
       })
     })
+    it(`should match operation in case of extension resource`, async () => {
```
jianyexi, line 561: could you add a test which can verify your change in "sway" ?

raych1: This test is for rpaas scenario, I added one more for arm scenario.

jianyexi (approved):

PR #607, opened 2021-04-06
oav2.3.1 dependency analyze. auto genereate test-scenario. run test scenario

ruowan:
scenario

### DIFF
--- a/lib/testScenario/gen/staticTestScenarioGenerator.ts
+++ b/lib/testScenario/gen/staticTestScenarioGenerator.ts
@@ -1,0 +1,151 @@
+import * as path from "path";
+import { inject, injectable } from "inversify";
+import { dump as yamlDump } from "js-yaml";
+import { JsonLoader } from "./../../swagger/jsonLoader";
+import { setDefaultOpts } from "./../../swagger/loader";
+
+import { FileLoader } from "./../../swagger/fileLoader";
+import { RawTestDefinitionFile, TestDefinitionFile } from "./../testResourceTypes";
+import { TestResourceLoaderOption } from "./../testResourceLoader";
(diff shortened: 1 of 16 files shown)


PhoenixHe-msft (approved):

lib/testScenario/gen/staticTestScenarioGenerator.ts
```diff
@@ -0,0 +1,151 @@
...
+          ),
+        },
+      ];
+      if (deleteExampleFilePath !== "") {
+        testStep.push({
+          step: deleteExampleName,
+          exampleFile: deleteExampleFilePath,
+        });
+      }
+      testDef.testScenarios.push({
```
raych1, line 102: Do we use same test scenario to store different PUT+DELETE cases when there're multiple examples for PUT operation?

lib/testScenario/swaggerAnalyzer.ts
```diff
@@ -0,0 +1,238 @@
...
+  filerTopLevelResourceType?: boolean;
+}
+
+export interface ExampleDependency {
+  exampleFilePath: string;
+  exampleName: string;
+  apiVersion: string;
+  resourceProvider: string;
+  resourceType: string;
+  internalDependency: Dependency;
```
raych1, line 40: Should this be array too?

ruowan: internalDependency is analyzed by path. So it should be only one dependencyChain.

lib/testScenario/gen/staticTestScenarioGenerator.ts
ruowan: No. For multiple examples, we store it in separrate test scenario

PR #618, opened 2021-05-11
examplegen support multiple operations

raych1:
support consuming multiple operations as input

### DIFF
--- a/lib/validate.ts
+++ b/lib/validate.ts
@@ -482,7 +482,7 @@ export async function extractXMsExamples(
 export async function generateExamples(
   specPath: string,
   payloadDir?: string,
-  operationId?: string,
+  operationIds?: string,
   readme?: string,
   tag?: string,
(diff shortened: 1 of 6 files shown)


lib/validate.ts
```diff
@@ -515,8 +515,11 @@ export async function generateExamples(
   log.filepath = options.logFilepath || log.filepath;
   for (const file of wholeInputFiles) {
     const generator = new ExampleGenerator(file, payloadDir);
-    if (operationId) {
-     await generator.generate(operationId);
+    if (operationIds) {
+     const operationIdArray = operationIds.trim().split(",");
+     for(const operationId of operationIdArray) {
```
jianyexi, line 520: do you need to remove empty operationId ? for example the input is  two commas

raych1: sounds good and ignore it.

PR #621, opened 2021-05-18
Supports different logging types in livevalidator

raych1:
Added loggingType to differentiate the log;
Added dedicated perf loggings.

### DIFF
--- a/lib/liveValidation/liveValidator.ts
+++ b/lib/liveValidation/liveValidator.ts
@@ -101,6 +101,13 @@ export enum LiveValidatorLoggingLevels {
   silly = "silly",
 }
 
+export enum LiveValidatorLoggingTypes {
+  trace = "trace",
+  perfTrace = "perfTrace",
+  error = "error",
+  incomingRequest = "incomingRequest",
+}
+
 /**
  * @class
  * Live Validator for Azure swagger APIs.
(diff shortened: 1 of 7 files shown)


lib/liveValidation/liveValidator.ts
```diff
@@ -196,25 +214,43 @@ export class LiveValidator {
       }
     }
 
-    this.logging("Transforming all specs.");
+    const startGlobalTransform = Date.now();
+    this.logging("Apply global transforms for all specs");
     this.loader.transformLoadedSpecs();
```
PhoenixHe-msft, line 219: No need to perf this one. This one is super fast.

lib/liveValidation/operationValidator.ts
```diff
@@ -75,11 +81,22 @@ export const validateSwaggerLiveRequest = async (
     }
     const startTimeToBuild = Date.now();
     validate = await loader.getRequestValidator(operation);
+    const elapsedTime = Date.now() - startTimeToBuild;
     if (logging) {
       logging(
-        `On-demand build request validator with DurationInMs:${Date.now() - startTimeToBuild}`,
+        `On-demand build request validator with DurationInMs:${elapsedTime}`,
```
PhoenixHe-msft, line 87: Is it possible to aggregate this metric in kusto?

PhoenixHe-msft (approved):

lib/liveValidation/operationValidator.ts
raych1: Yes, we can.

lib/liveValidation/liveValidator.ts
raych1: Removed it.

PR #634, opened 2021-06-25
support extract output from armTemplate deployment

ruowan:


lib/testScenario/postmanCollectionRunnerClient.ts
```diff
@@ -371,6 +371,7 @@ export class PostmanCollectionRunnerClient implements TestScenarioRunnerClient {
       raw: JSON.stringify(body, null, 2),
     });
     this.addAuthorizationHeader(item);
+    console.log(armTemplate.outputs);
```
raych1: Will output include sensitive data?

lib/testScenario/postmanCollectionRunnerClient.ts
```diff
@@ -341,13 +343,11 @@ export class PostmanCollectionRunnerClient implements TestScenarioRunnerClient {
     this.auth(stepEnv.env);
     const item = new Item();
     item.name = step.step;
-    const path =
-      "/subscriptions/:subscriptionId/resourcegroups/:resourceGroupName/providers/Microsoft.Resources/deployments/{{deploymentName}}?api-version=2020-06-01";
```
raych1, line 345: where does `deploymentName` come from before?

lib/testScenario/postmanCollectionRunnerClient.ts
ruowan: Updated. It's just a debug log. I remove it.

lib/testScenario/postmanCollectionRunnerClient.ts
ruowan: From env. but env is not define. So the value is undefine. It's a bug before.

zhenglaizhang (approved):

raych1 (approved):

PR #636, opened 2021-06-25
Bump deps

zhenglaizhang:


raych1 (approved):

raych1:
snapshots needs to be patched due to package.json change.

PR #637, opened 2021-07-05
Ignore LRO header check in case of sync calls

raych1:
If the api call is sync call, then the response header won't have LRO headers.


lib/liveValidation/operationValidator.ts
```diff
@@ -358,9 +359,25 @@ const validateLroOperation = (
 
 const validateLroHeader = (
   operation: Operation,
+  statusCode: string,
   headers: StringMap<string>,
+  body: StringMap<unknown> | undefined,
```
PhoenixHe-msft, line 364: body should be type of any

PhoenixHe-msft (approved):

PR #642, opened 2021-07-15
Bug Fix: validate-traffic command doesn't support relative path input

Ligengxin96:


Ligengxin96:
### Absolute path
image


### Relative path
image

raych1 (approved):

PR #644, opened 2021-07-21
Update readme and small bug fix

ruowan:


lib/testScenario/bodyTransformer.ts
```diff
@@ -101,9 +101,9 @@ export class BodyTransformer {
 
     if (typeof dst === "object" && typeof src === "object") {
       const result: any = { ...dst };
```
jianyexi, line 103: what if dst is null and src is not null ?

ruowan: It will not entry this code. So it's safe

jianyexi: could you add comment here

ChangeLog.md
```diff
@@ -5,6 +5,8 @@
 - Oav runner support specifying option 'runId' with option 'from' and/or 'to' to debug.
 - Oav runner support using env variable 'TEST_SCENARIO_JSON_ENV' to override variables in env.json.
 - Replace Oav runner option 'cleanUp' with 'skipCleanUp'.
+- Update README.md add gif to show API test
```
raych1, line 8: Why not bump a new version for these new changes?

README.md
```diff
@@ -17,17 +16,39 @@ You can install the latest stable release of node.js from [here](https://nodejs.
...
 ```bash
-npm install -g oav
+npm install -g oav@latest
 ```
 
+![](./documentation/installOav.gif)
+
+### Run API test
+
+OAV support run API test against Azure and validate request and response. You could define test scenario file which compose with serval swagger example file and then use oav to run it. For more details about API test, please refer to this [API testing doc](https://github.com/Azure/azure-rest-api-specs/tree/test-scenario-main/documentation/test-scenario).
```
raych1, line 26: Typo of `several`.

ChangeLog.md
ruowan: 2.7.0 is not released yet.

lib/testScenario/reportGenerator.ts
```diff
@@ -420,9 +420,8 @@ export class ReportGenerator {
         })
         .filter((it) => it !== undefined) as ResponseDiffItem[];
       return delta;
-    } catch (err) {
-      console.log(err);
```
raych1, line 420: Do you mean log will cause exception?

ruowan: Yes. Do not output error message here. It make confuse for customer

raych1 (approved):

PR #646, opened 2021-08-03
Disable checkUnderFileRoot in `Validate Traffic` command

Ligengxin96:


Ligengxin96:
As we discussed in the meeting, disable checkUnderFileRoot to suppress  the `load file outside of root folder` error.

raych1 (approved):

PR #649, opened 2021-08-09
Refactor Semantic Feature

PhoenixHe-TSP:


lib/liveValidation/liveValidator.ts
```diff
@@ -70,11 +70,11 @@ export interface RequestResponseLiveValidationResult {
   readonly runtimeException?: RuntimeException;
 }
 
-export interface LiveValidationIssue extends SchemaValidateIssue {
-  readonly pathsInPayload: string[];
-  readonly severity: Severity;
-  readonly documentationUrl?: string;
-}
+export type LiveValidationIssue = {
```
raych1, line 73: Is severity not used?

lib/liveValidation/liveValidator.ts
```diff
@@ -198,7 +198,7 @@ export class LiveValidator {
       container,
       fileRoot: this.options.directory,
       ...this.options,
-      loadSuppression: this.options.loadSuppression ?? Object.keys(allErrorConstants),
+      loadSuppression: this.options.loadSuppression ?? Object.keys(),
```
raych1, line 201: Why do we change default behavior?

PhoenixHe-msft: Unfinished. Should use another way to compute all the errorcode

lib/liveValidation/liveValidator.ts
PhoenixHe-msft: It should move to lower level.

lib/validate.ts
```diff
@@ -122,26 +121,8 @@ const prettyPrint = <T extends NodeError<T>>(
   }
 };
 
-export const validateSpec = async (
-  specPath: string,
-  options: Options | undefined
-): Promise<SpecValidationResult> =>
+export const validateSpec = async (specPath: string, options: Options | undefined) =>
   validate(options, async (o) => {
```
raych1, line 125: Do we keep these options?

raych1 (approved):

PR #655, opened 2021-08-17
V2

leni-msft:


lib/autorestPlugin/extension.ts
```diff
@@ -54,7 +54,7 @@ async function analyzeSwagger(
   autoRestApi: extensionBase.Host
 ): Promise<void> {
   const swaggerFile = await autoRestApi.ReadFile(swaggerFileName);
-  const swagger = yaml.safeLoad(swaggerFile);
+  const swagger = yaml.load(swaggerFile);
```
raych1, line 57: Is the old api deprecated?

lib/swagger/swaggerTypes.ts
```diff
@@ -228,6 +228,7 @@ export interface Schema extends BaseSchema {
   oneOf?: Schema[];
   additionalProperties?: boolean | Schema;
   properties?: { [propertyName: string]: Schema };
+  patternProperties?: { [propertyPattern: string]: Schema };
```
raych1, line 231: What is this used for?

lib/autorestPlugin/extension.ts
leni-msft: yes, it's supposed to use `load` in the new version of js-yaml

lib/swagger/swaggerTypes.ts
leni-msft: It's part of JSON schema, to specify schema for certain pattern properties in Json Schema. https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties

lib/commands/generate-test-static-scenario.ts
```diff
@@ -41,7 +41,7 @@ export async function handler(argv: yargs.Arguments): Promise<void> {
   console.log("input-file:");
   console.log(swaggerFilePaths);
 
-  const generator = StaticTestScenarioGenerator.create({
```
ruowan, line 44: Maybe we should wait for final product name?

lib/testScenario/gen/testScenarioGenerator.ts
```diff
@@ -50,21 +46,21 @@ export interface RequestTracking {
   description: string;
 }
 
-export interface TestScenarioGeneratorOption extends TestResourceLoaderOption {}
+export interface TestScenarioGeneratorOption extends ApiScenarioLoaderOption {}
```
ruowan, line 49: This file is used for generate test scenario from payload. I think later we should merge this functionationality with static test scenario generation just in one command.

lib/testScenario/apiScenarioYamlLoader.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  public async load(filePath: string): Promise<RawScenarioDefinition> {
+    this.fileCache.clear();
+
+    const fileContent = await this.fileLoader.load(filePath);
+    yamlLoad(fileContent, {
+      schema: DEFAULT_SCHEMA.extend(
+        new Type("!include", {
+          kind: "scalar",
```
ruowan, line 31: what does the kind 'scalar' mean? better to add doc link about howt to add customize extension for yaml.

leni-msft:
Docs PR: https://github.com/Azure/azure-rest-api-specs/pull/16086

lib/testScenario/gen/testScenarioGenerator.ts
leni-msft: Per discussion offline, we'll do the renaming in this branch.

lib/testScenario/apiScenarioYamlLoader.ts
leni-msft: It's data type of YAML standard, not supposed to be used by external users.

lib/commands/generate-test-static-scenario.ts
leni-msft: We'd like to make all the required breaking changes in one PR.

lib/xMsExampleExtractor.ts
```diff
@@ -221,11 +221,20 @@ export class XMsExampleExtractor {
                     bodyParamValue !== "" ? JSON.parse(bodyParamValue) : "";
                 }
               }
+
+              const parseResponseBody = (body: any) => {
+                try {
+                  return JSON.parse(body);
+                } catch (err) {
+                  return body;
```
raych1, line 229: Can you log error saying the body is mal-formatted? The example generated is not a valid file.

ruowan: It's extract example from recording and write back example to swagger. So the response payload has some non-json format such as xml. 

We don't log error here. Since it's a trial for parsing json.

raych1 (approved):

PR #686, opened 2021-10-18
update tests for semantic validator

ankhyk:


test/semanticValidatorTests.ts
```diff
@@ -8,51 +8,205 @@ import * as validate from "../lib/validate";
...
+      assert(result.validityStatus === false);
+      assert.strictEqual(
+        result.validateSpec?.errors?.[0].code,
+        "MISSING_PATH_PARAMETER_DEFINITION"
+      );
+    });
   });
 });
+
+// it("should validate correctly when the spec contains an x-ms-parameterized-host", async () => {
```
raych1, line 181: Are all of below tests not valid?

ankhyk: just the third test("should succeed when discriminator is not a required property and the error is suppressed") is not valid

these are the original test codes
should I put back the ones that work?

PR #693, opened 2021-10-25
Ruowan/gen api scenario

ruowan:


lib/commands/generate-static-api-scenario.ts
```diff
@@ -19,31 +19,43 @@ export const builder: yargs.CommandBuilder = {
...
-  const readmeMd: string = pathResolve(argv.readme);
-
-  const autorestConfig = await getAutorestConfig(argv, readmeMd);
-  const fileRoot = dirname(readmeMd);
-  const swaggerFilePaths: string[] = autorestConfig["input-file"].map((it: string) =>
-    pathResolve(fileRoot, it)
-  );
+  const swaggerFilePaths: string[] = (argv.swaggers || []).map((it: string) => pathResolve(it));
+  let tag = "default";
+  if (argv.readme !== undefined) {
```
raych1, line 38: I assume it supports scenario file generation when both of swaggers parameter and readme parameter are passed in, right?

lib/apiScenario/gen/staticTestScenarioGenerator.ts
```diff
@@ -209,10 +209,11 @@ export class StaticApiScenarioGenerator {
     for (const it of this.exampleDependencies) {
       resourceTypes.add(it.fullResourceType);
     }
+    if (this.opts.rules?.includes("operations-list")) {
+      console.log("generate operations list");
+      await this.generateListOperationTestScenario();
```
raych1, line 214: Do we change generation behavior with moving this call outside of resourceTypes loop?

ruowan: No

lib/commands/generate-static-api-scenario.ts
ruowan: Yes.

raych1 (approved):

PR #694, opened 2021-10-26
Fix error conversion issue in MV

raych1:
It reports invalid token reference when parsing the error of 'invalid_type' in case of null body in response.


ruowan (approved):

ruowan:
pipeline failed

jianyexi (approved):

PR #695, opened 2021-10-27
bug fix for api scenario generation

ruowan:


lib/apiScenario/gen/staticTestScenarioGenerator.ts
```diff
@@ -68,6 +68,10 @@ export class StaticApiScenarioGenerator {
     const paths = await this.swaggerAnalyzer.getOperationListPath();
     for (const it of paths) {
       const example = Object.values(it.get?.["x-ms-examples"] || {});
+      if (example[0] === undefined) {
+        console.log("can not found example for operations-list");
```
raych1, line 72: Can we also output the api path info?

raych1 (approved):

lib/apiScenario/gen/staticTestScenarioGenerator.ts
ruowan: api path is operations  list.

raych1: No, I mean the swagger info. Otherwise, we don't know which one miss the operation_list.

PR #696, opened 2021-10-28
support dataplane traffic validation

raych1:
support multiple swaggers and multiple traffic files validation;
support path parameter transform in live validation and example validation.


lib/liveValidation/liveValidator.ts
```diff
@@ -850,7 +850,7 @@ export const parseValidationRequest = (
     // Provider would be provider found from the path or Microsoft.Unknown
     providerNamespace = providerNamespace || C.unknownResourceProvider;
     if (providerNamespace === C.unknownResourceProvider) {
-      apiVersion = C.unknownApiVersion;
+      //apiVersion = C.unknownApiVersion;
```
scbedd, line 853: :) I see why you're asking now 👍

lib/liveValidation/operationValidator.ts
```diff
@@ -99,6 +99,7 @@ export const validateSwaggerLiveRequest = async (
   }
 
   const pathParam = extractPathParamValue(info.operationMatch!);
+  transformMapValue(pathParam, operation._pathTransform);
```
ruowan, line 102: what does this do?

lib/liveValidation/operationSearcher.ts
```diff
@@ -154,7 +154,7 @@ export class OperationSearcher {
         undefined,
         requestInfo
       );
-      requestInfo.apiVersion = unknownApiVersion;
+      //requestInfo.apiVersion = unknownApiVersion;
```
ruowan, line 157: why comment here

lib/liveValidation/liveValidator.ts
raych1: you got it~

lib/liveValidation/operationSearcher.ts
raych1: use matched api version instead.

lib/liveValidation/operationValidator.ts
raych1: transform path parameters value per types. e.g. 'false' - > false.

ruowan (approved):

PR #698, opened 2021-11-04
check discriminator property type must be string

jordan13game:
If a property is declared as discriminator, the property type must be string and nothing else


raych1:
Code looks good. Can you bump the patch version and add changelog? Also update [error document](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/Semantic-and-Model-Violations-Reference.md) in another PR.

raych1 (approved):

PR #702, opened 2021-11-08
fix jsonPathToArray bug and update snapshot

ankhyk:


regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -160,8 +160,8 @@ Array [
     "code": "INVALID_FORMAT",
     "exampleJsonPath": "$parameters.serviceMemberId",
     "examplePosition": Object {
-      "column": 17,
-      "line": 2,
+      "column": 24,
```
raych1, line 163: Did you sample check if column/line no are correct?

ankhyk: I checked the first few and they are all correct

raych1:
can you add changelog?

raych1 (approved):

PR #704, opened 2021-11-10
Fix 'generate-examples' command failed to generate correct format for ‘uri’ format parameter

Ligengxin96:


lib/generator/mocker.ts
```diff
@@ -48,7 +48,14 @@ export default class Mocker {
       );
     }
     const length = this.getRandomInt(minLength, maxLength);
-    return "a".repeat(length);
+    let mockedValue = "a".repeat(length);
+
+    if (paramSpec.format === "uri") {
+      const prefix = "https://a";
```
raych1: Can you add hostname like https://host.com/a? Also pls guarantee the generated value passes model validation.

raych1 (commented):
Can you check if the spec validated in exampleGeneratorTests cover url scenario? Please add it if it's not covered.

test/__snapshots__/exampleGeneratorTests.ts.snap
```diff
@@ -125,6 +125,85 @@ Array [
...
   ],
+  Object {
+    "code": "INVALID_FORMAT",
+    "exampleJsonPath": "$parameters.parameters",
+    "examplePosition": Object {
+      "column": 19,
+      "line": 8,
+    },
+    "exampleUrl": "test/exampleGenerator/specification/sql/resource-manager/Microsoft.Sql/preview/2020-02-02-preview/examples/ManagedDatabaseCompleteExternalRestore.json",
+    "message": "Object didn't pass validation for format uri: testdb1_log4",
```
raych1: Can you share why test still doesn't pass URL format validation?

Ligengxin96: Oh, I forgot update the example file.

raych1 (approved):

PR #705, opened 2021-11-15
Make response headers as optional

raych1:
It's base on swagger 2.0 specification.

### DIFF
--- a/lib/liveValidation/operationValidator.ts
+++ b/lib/liveValidation/operationValidator.ts
@@ -170,7 +170,7 @@ export const validateSwaggerLiveResponse = async (
   if (rsp.schema !== undefined) {
     validateContentType(operation.produces!, headers, false, result);
     if (isArmCall && realCode >= 200 && realCode < 300) {
-      validateLroOperation(operation, statusCode, headers, body, result);
+      validateLroOperation(operation, statusCode, headers, result);
     }
   }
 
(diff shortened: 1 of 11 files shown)


ruowan (approved):

lib/liveValidation/operationValidator.ts
```diff
@@ -358,21 +357,11 @@ const validateLroHeader = (
...
-      (properties.provisioningState === undefined ||
-        properties.provisioningState === "Succeeded" ||
-        properties.provisioningState === "Failed" ||
-        properties.provisioningState === "Canceled")
-    ) {
-      // Ignore LRO header check when it's sync call
-      return;
-    }
+  if (statusCode === "201") {
+    // Ignore LRO header check cause RPC says azure-AsyncOperation is optional if using 201/200+ provisioningState
```
ruowan, line 363: What if statusCode is  202? Do we validate 202 header?

lib/liveValidation/operationValidator.ts
```diff
@@ -322,31 +322,30 @@ const validateLroOperation = (
   operation: Operation,
   statusCode: string,
   headers: StringMap<string>,
-  body: any,
```
ruowan, line 325: Is body unused?

lib/liveValidation/operationValidator.ts
raych1: Yes, 202 case has to provide LRO headers.

lib/liveValidation/operationValidator.ts
raych1: Yes, no need any more.

PR #706, opened 2021-11-16
fix url encoding issue

ruowan:


lib/util/utils.ts
```diff
@@ -489,6 +489,15 @@ export function isUrlEncoded(str: string): boolean {
   }
 }
 
+export function paramsToObject(entries: any) {
```
raych1: Can you reanme it more readable? like kvPairsToObject?

lib/liveValidation/liveValidator.ts
```diff
@@ -834,15 +838,15 @@ export const parseValidationRequest = (
...
   if (pathStr !== "") {
     // Lower all the keys and values of query parameters before searching for `api-version`
-    const queryObject = _.transform(
-      parsedUrl.query,
-      (obj: ParsedUrlQuery, value, key) =>
-        (obj[key.toLowerCase()] = _.isString(value) ? value.toLowerCase() : value)
-    );
+    const queryObject: ParsedUrlQuery = {};
+    parsedUrl.searchParams.forEach((value, key) => {
+      queryObject[key.toLowerCase()] = value;
```
raych1: Should value be lowercased when it's string?

ruowan: Key must be string

lib/liveValidation/operationSearcher.ts
```diff
@@ -343,11 +343,22 @@ export function getProviderFromPathTemplate(pathStr?: string | null): string | u
...
+  const manageManagementMatch = managementPlaneProviderInSpecPathRegEx.exec(specPath);
+  const dataPlaneMatch = dataPlaneProviderInSpecPathRegEx.exec(specPath);
+  return manageManagementMatch === null
+    ? dataPlaneMatch === null
+      ? undefined
+      : { provider: dataPlaneMatch[1], type: "data-plane" }
+    : { provider: manageManagementMatch[1], type: "resource-manager" };
+}
+const managementPlaneProviderInSpecPathRegEx = new RegExp("/(resource-manager)/(:?[^{/]+)", "gi");
+const dataPlaneProviderInSpecPathRegEx = new RegExp("/(resource-manager)/(:?[^{/]+)", "gi");
```
raych1: Should be data-plane in regex.

lib/templates/httpTemplate.ts
```diff
@@ -30,7 +30,9 @@ export class HttpTemplate {
 
   protected getHost(): string | undefined {
     const requestUrl = this.request.url;
-    return requestUrl ? url.parse(requestUrl).host : "management.azure.com";
+    return requestUrl
+      ? new URL(requestUrl, "https://management.azure.com").host
```
raych1, line 34: Is this default domain when it's absent?

test/liveValidatorDataPlaneTests.ts
```diff
@@ -0,0 +1,28 @@
...
+      const tableSwaggerFilePath =
+        "cosmos-db/data-plane/Microsoft.Tables/preview/2019-02-02/table.json";
+
+      const options = {
+        swaggerPathsPattern: [tableSwaggerFilePath],
+        directory: "./test/liveValidation/swaggers/specification",
+      };
+      const liveValidator = new LiveValidator(options);
+      await liveValidator.initialize();
+      expect(liveValidator.operationSearcher.cache.size).toEqual(1);
```
raych1, line 20: Can you check if the key equals unknown? Remove below console line if for debugging.

test/liveValidatorDataPlaneTests.ts
```diff
@@ -0,0 +1,28 @@
...
+      console.log(liveValidator.operationSearcher.cache.keys());
+    });
+
+    it("should initialize multi data-plane swagger without conflict", async () => {
```
raych1: Can you skip this incomplete test?

test/liveValidation/swaggers/specification/cosmos-db/data-plane/readme.go.md
```diff
@@ -0,0 +1,26 @@
+## Go
```
raych1: Can you remove all readme files which are not requried?

lib/liveValidation/operationSearcher.ts
```diff
@@ -343,11 +343,22 @@ export function getProviderFromPathTemplate(pathStr?: string | null): string | u
...
-export function getProviderFromSpecPath(specPath: string): string | undefined {
-  const match = providerInSpecPathRegEx.exec(specPath);
-  return match === null ? undefined : match[1];
+export interface PathProvider {
+  provider: string;
+  type: "resource-manager" | "data-plane";
 }
-const providerInSpecPathRegEx = new RegExp("/resource-manager/(:?[^{/]+)", "gi");
+
+export function getProviderFromSpecPath(specPath: string): PathProvider | undefined {
```
weshaggard, line 351: @zhenglaizhang this looks like a place where we make some assumptions about the path structure and conventions in the specs repo. Do we know how many places we are doing this similar processing across all our various tools? We probably need to extract this logic out into one tool or script so that we can avoid this duplication.

weshaggard:
@ruowan can you link to the issue you are trying to fix with this PR so I can see a little more context? (thanks for adding me as a code reviewer).

lib/liveValidation/liveValidator.ts
raych1: I meant `value` that could be apiversion like 2020-01-01-PREview.

lib/liveValidation/operationSearcher.ts
raych1: Thanks @weshaggard for the suggestion. There're a few places describing spec repo path pattern but not all of them are same. We will need to collect them and revisit whether they could be consolidated.

ruowan:
Hi, @weshaggard This work item. https://devdiv.visualstudio.com/DevDiv/_sprints/taskboard/Azure%20Swagger%20Tools/DevDiv/OneVS/Nickel%20%5BNI%5D/Q4/2021-11?workitem=1429165

lib/templates/httpTemplate.ts
ruowan: yes

raych1:
@ruowan , can you remove all the examples and readme.md under `Microsoft.Tables` since they're not used?

lib/liveValidation/liveValidator.ts
ruowan: Good catch

ruowan:
I think to make the whole swagger integrity. We don't need to remove them

lib/liveValidation/liveValidator.ts
```diff
@@ -844,7 +844,7 @@ export const parseValidationRequest = (
     // Lower all the keys and values of query parameters before searching for `api-version`
     const queryObject: ParsedUrlQuery = {};
     parsedUrl.searchParams.forEach((value, key) => {
-      queryObject[key.toLowerCase()] = value;
+      queryObject[key.toLowerCase()] = value.toLowerCase();
```
raych1, line 847: Can you check type of value before calling `tolowercase`?  
```
_.isString(value) ? value.toLowerCase() : value
```

ruowan: Search Parameter value must be string. We do not need additional check.

ruowan: ```    forEach(callback: (value: string, name: string, searchParams: this) => void): void;``` method signiture.

raych1 (approved):

PR #707, opened 2021-11-17
add new rules of discriminator

jordan13game:
add new rules of 'INVALID_DISCRIMINATOR_VALUE' and 'DISCRIMINATOR_PROPERTY_NOT_FOUND'


lib/swaggerValidator/semanticValidator.ts
```diff
@@ -346,12 +347,49 @@ export class SwaggerSemanticValidator {
...
-          property: d,
-        });
-        this.addErrorsFromErrorCode(errors, rootInfo.url, meta, sch);
+      } else {
+        const discriminatorProp = jsonLoader.resolveRefObj(sch.properties[d]);
+        if (discriminatorProp.type !== "string") {
+          const meta = getOavErrorMeta("INVALID_DISCRIMINATOR_TYPE", {
+            property: d,
+          });
+          this.addErrorsFromErrorCode(errors, rootInfo.url, meta, sch);
```
ruowan, line 359: Do we still need validate further, if already find INVALID_DISCRIMINATOR_TYPE error.

lib/swagger/swaggerTypes.ts
```diff
@@ -260,6 +260,9 @@ export interface Schema extends BaseSchema {
   [refSelfSymbol]?: string;
 
   _skipError?: boolean;
+
+  // x-ms-discriminator-value exists but discriminator missing
+  _missingDiscriminator?: boolean;
```
ruowan, line 265: Why record status in swagger type?

jordan13game: Because if a schema defined x-ms-discriminator-value but its parent doesn't define discriminator, it won't be record in discriminatorMap, so we can't find it in discriminatorMap. And then in semantic validation, the 'allOf' field is deleted, so we also don't know which is inherited from. So we record the information in the transform.

lib/util/errorDefinitions.ts
```diff
@@ -254,11 +254,21 @@ export const semanticValidationErrors = {
     message: strTemplate`Discriminator must be a required property.`,
     id: "OAV131",
   },
-  DISCRIMINATOR_PROPERTY_TYPE_NOT_STRING: {
+  INVALID_DISCRIMINATOR_TYPE: {
     severity: Severity.Critical,
     message: strTemplate`The property type of discriminator must be string.`,
     id: "OAV132",
   },
+  INVALID_DISCRIMINATOR_VALUE: {
```
raych1: How about rename error code to `INVALID_XMS_DISCRIMINATOR_VALUE`?

lib/util/errorDefinitions.ts
```diff
@@ -254,11 +254,21 @@ export const semanticValidationErrors = {
...
+    severity: Severity.Critical,
+    message: strTemplate`The value of x-ms-dicriminator-value is not in the discriminator enum list: ${"value"}`,
+    id: "OAV133",
+  },
+  DISCRIMINATOR_PROPERTY_NOT_FOUND: {
+    severity: Severity.Critical,
+    message: strTemplate`Discriminator is missing in the parent of the x-ms-dicriminator-value: ${"value"}`,
```
raych1: Let's use message: `Missing discriminator in base model. This derived model has x-ms-dicriminator-value: ${"value"}`.

ChangeLog.md
```diff
@@ -6,6 +6,7 @@
 - LiveValidatorLoader - make response headers as optional for swagger 2.0
 - LiveValidator - ignore LRO header check if using 201/200+ provisioningState
 - ModelValidator/SemanticValidator - support suppression on error code
+- Add new rules of 'INVALID_DISCRIMINATOR_VALUE' and 'DISCRIMINATOR_PROPERTY_NOT_FOUND'
```
raych1: Can you add prefix of `semantic validation - `?

lib/swaggerValidator/semanticValidator.ts
```diff
@@ -349,12 +350,48 @@ export class SwaggerSemanticValidator {
...
         this.addErrorsFromErrorCode(errors, rootInfo.url, meta, sch);
-      } else if (sch.properties[d].type !== "string") {
-        const meta = getOavErrorMeta("DISCRIMINATOR_PROPERTY_TYPE_NOT_STRING", {
-          property: d,
-        });
-        this.addErrorsFromErrorCode(errors, rootInfo.url, meta, sch);
+      } else {
+        const discriminatorProp = jsonLoader.resolveRefObj(sch.properties[d]);
+        if (discriminatorProp.type !== "string") {
+          const meta = getOavErrorMeta("INVALID_DISCRIMINATOR_TYPE", {});
```
raych1: Can you output the property name in the error message?

raych1 (approved):

PR #718, opened 2021-12-06
update mv tests

ankhyk:


ankhyk:
oav-merge-validation will report error when: 
- when missing $ref in x-ms-examples
- when specPath is a url starting with https://github

raych1 (commented):
Can you add changeLog?

raych1 (approved):

PR #726, opened 2021-12-30
Fix false alarm for WRITEONLY_PROPERTY_NOT_ALLOWED_IN_RESPONSE

ankhyk:
to ignore the false alarm
https://portal.azure-devex-tools.com/amekpis/correctness/detail?errorId=70cc8080-d828-4cdb-87cd-1376c63c5074

### DIFF
--- a/lib/swaggerValidator/ajvSchemaValidator.ts
+++ b/lib/swaggerValidator/ajvSchemaValidator.ts
@@ -219,10 +219,9 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
   if (
     cxt.isResponse &&
     (cxt as any)?.httpMethod === "post" &&
-    // if keyword is x-ms-secret, required check can be ignored when x-ms-secret is "true"
+    // should skip error when x-ms-secret is "true"
     ((keyword === "x-ms-secret" && (parentSchema as any)?.[xmsSecret] === true) ||
-      // if keyword is x-ms-mutability,
-      // required check can be ignored when x-ms-secret is "true" and x-ms-mutability is "create" and "update"
+      // should skip error when x-ms-secret is "true" and x-ms-mutability is "create" and "update"
       (keyword === "x-ms-mutability" &&
         (parentSchema as any)?.[xmsSecret] === true &&
         parentSchema[xmsMutability]?.indexOf("read") === -1))
--- a/lib/swaggerValidator/ajvSchemaValidator.ts
+++ b/lib/swaggerValidator/ajvSchemaValidator.ts
@@ -218,8 +218,9 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
   // If a response has property which x-ms-secret value is true in post we can skip this error
   if (
     cxt.isResponse &&
-    (parentSchema.properties?.[(params as any).missingProperty] as any)?.[xmsSecret] === true &&
-    cxt.httpMethod === "post"
+    (cxt as any)?.httpMethod === "post" &&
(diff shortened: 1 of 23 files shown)


lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -215,6 +215,16 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
     return true;
   }
 
+  // If a response has property which x-ms-secret value is true in post we can skip this error
+  if (
```
raych1, line 219: Can you check if traffic validation will fail when the schema only has "x-ms-mutability": ["create","update"]?

raych1 (commented):
Can you add more tests for x-ms-mutability?

test/liveValidatorTests.ts
```diff
@@ -744,19 +744,84 @@ describe("Live Validator", () => {
...
+      const liveValidator = new LiveValidator(options);
+      await liveValidator.initialize();
+      const payload = require(`${__dirname}/liveValidation/payloads/xmsSecretAndPOST/xmsSecretAndPOST_2021-03-01.json`);
+      const result = await liveValidator.validateLiveRequestResponse(payload);
+      assert.equal(result.responseValidationResult.isSuccessful, false);
+    });
+
+    it(`should report error in response for httpMethod which is not POST,
+    when x-ms-secret is "true" or x-ms-mutability is "create" and "update"`, async () => {
+      const swaggers = [
```
raych1: What's the difference between these three swaggers? can you add the main difference in the comment for better understanding?

ankhyk: I've added comments in test
The general structure is as follows

- [√]httpMethod: POST, [√]x-ms-secret: "true", [X]x-ms-mutability: "create" and "update"
- [√]httpMethod: POST, [√]x-ms-secret: "true", [√]x-ms-mutability: "create" and "update"
- [√]httpMethod: POST, [X]x-ms-secret: "true", [√]x-ms-mutability: "create" and "update"
- [X]httpMethod: POST
  - [√]x-ms-secret: "true", [X]x-ms-mutability: "create" and "update"
  - [√]x-ms-secret: "true", [√]x-ms-mutability: "create" and "update"
  - [X]x-ms-secret: "true", [√]x-ms-mutability: "create" and "update"

test/liveValidatorTests.ts
```diff
@@ -738,88 +738,97 @@ describe("Live Validator", () => {
...
+            git: {
+              shouldClone: false,
+            },
+          };
+          const liveValidator = new LiveValidator(options);
+          await liveValidator.initialize();
+          const payloadVersion = swagger.slice(0, -7);
+          const payload = require(`${__dirname}/liveValidation/payloads/xmsSecretAndPOST/xmsSecretButGet_${payloadVersion}.json`);
+          const result = await liveValidator.validateLiveRequestResponse(payload);
+          assert.equal(result.responseValidationResult.isSuccessful, false);
```
raych1, line 838: It needs to further validate if the error code is expected.

test/liveValidatorTests.ts
```diff
@@ -838,8 +839,9 @@ describe("Live Validator", () => {
 					const errors = result.responseValidationResult.errors;
 					for (const error of errors) {
 						assert.equal(
-							error.code === "WRITEONLY_PROPERTY_NOT_ALLOWED_IN_RESPONSE" ||
-								error.code === "SECRET_PROPERTY",
+							(error.schemaPath.indexOf("x-ms-secret") !== -1 && error.code === "SECRET_PROPERTY") ||
```
raych1, line 842: Indent issue?

raych1 (approved):

PR #744, opened 2022-01-20
loosen date-time validation

ankhyk:
should not report error when date-time value is just missing a Z in the end


lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -229,6 +229,23 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
     return true;
   }
 
+  // If a response has property which x-ms-secret value is "true" in post we can skip this error
```
raych1: Comment needs to be changed.

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -229,6 +229,23 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
...
+  if (
+    cxt.isResponse &&
+    keyword === "format" &&
+    schema === "date-time" &&
```
raych1: Can you confirm if schema will have the value of 'date-time-rfc1123'?

test/modelValidation/swaggers/specification/dateTime/dateTime.json
```diff
@@ -0,0 +1,259 @@
...
+          },
+          "type": "string",
+          "description": "State of user session."
+        },
+        "activeDirectoryUserName": {
+          "type": "string",
+          "description": "The active directory user name."
+        },
+        "createTime": {
+          "format": "date-time",
```
raych1: Can you have another case of 'date-time-rfc1123'?

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -229,6 +229,26 @@ const shouldSkipError = (error: ErrorObject, cxt: SchemaValidateContext) => {
     return true;
   }
 
+  // If payload has property with date-time parameter and its value is valid except missing "Z" in the end we can skip this error
+  if (
+    cxt.isResponse &&
```
raych1: `cxt.isResponse` can also be removed coz the check applies for request too.

raych1 (commented):
Can you modify the swagger and payload to cover `datetime` property validation in request either?

ankhyk:
> Can you modify the swagger and payload to cover `datetime` property validation in request either?

The payload already has date-time property in both request and response

raych1 (approved):

PR #757, opened 2022-02-23
Xiaoxu/coverage: add coverage for each swagger file for trafficValidator

xiaoxuqi-ms:


lib/swaggerValidator/trafficValidator.ts
```diff
@@ -25,6 +26,8 @@ export class TrafficValidator {
   private trafficFiles: string[] = [];
   private specPath: string;
   private trafficPath: string;
+  private trafficOperation: Map<string, string[]> = new Map<string, string[]>();
+  public coverageResult: Map<string, number> = new Map<string, number>();
```
raych1: As discussed, let's return array of JSON like [{"spec":"", "coveredOperations": number, "totalOperations": number, "coverageRate": number }]

lib/liveValidation/liveValidator.ts
```diff
@@ -174,6 +177,7 @@ export class LiveValidator {
     this.options = ops as LiveValidatorOptions;
     this.logging(`Creating livevalidator with options:${JSON.stringify(this.options)}`);
     this.operationSearcher = new OperationSearcher(this.logging);
+    this.operationSpecMapper = new Map<string, string[]>();
```
raych1: How about moving this information to trafficValidation class as it's not useful here?

lib/liveValidation/liveValidator.ts
```diff
@@ -217,8 +219,10 @@ export class LiveValidator {
     );
 
     const allSpecs: SwaggerSpec[] = [];
+    this.swaggerList = [];
```
raych1: Can you move initialization to declaration place?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -19,12 +25,24 @@ export interface RuntimeException {
   message: string;
 }
 
+export interface CoverageInfo {
```
raych1: How about rename to `OperationCoverageInfo`?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -67,6 +85,44 @@ export class TrafficValidator {
...
+
+    const container = inversifyGetContainer();
+    this.loader = inversifyGetInstance(LiveValidatorLoader, {
+      container,
+      fileRoot: liveValidationOptions.directory,
+      ...liveValidationOptions,
+      loadSuppression: Object.keys(apiValidationErrors),
+    });
+
+    // re-set the transform context after set the logging function
```
raych1: line#97 and line#98 are not needed here. They're used to reset the logger.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -67,6 +85,44 @@ export class TrafficValidator {
...
+      let spec;
+      try {
+        spec = await this.loader.load(pathResolve(swaggerPath));
+      } catch (e) {
+        console.log(e);
+      }
+      if (spec !== undefined) {
+        // Get Swagger - operation mapper.
+        if (this.operationSpecMapper.get(swaggerPath) === undefined) {
+          this.operationSpecMapper.set(swaggerPath, []);
```
raych1, line 119: lowercase the swaggerPath here when add to map then fetch doesn't need to do it again.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -111,6 +177,44 @@ export class TrafficValidator {
...
+    });
     return this.trafficValidationResult;
   }
+
+  private findSwaggerByOperationInfo(operationInfo: OperationContext) {
+    let result = undefined;
+    this.operationSpecMapper.forEach((value: string[], key: string) => {
+      if (
+        toLower(key).includes(toLower(operationInfo.apiVersion)) &&
+        toLower(key).includes(toLower(operationInfo.validationRequest?.providerNamespace))
```
raych1: return if validationRequest is undefined.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -111,6 +177,44 @@ export class TrafficValidator {
...
+      ) {
+        if (value.includes(operationInfo.operationId)) {
+          result = key;
```
raych1, line 262: break the loop when find result.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -77,6 +133,16 @@ export class TrafficValidator {
         const payload = require(trafficFile);
         const validationResult = await this.liveValidator.validateLiveRequestResponse(payload);
         const operationInfo = validationResult.requestValidationResult?.operationInfo;
+        const swaggerFile = this.findSwaggerByOperationInfo(operationInfo);
+        if (swaggerFile !== undefined) {
```
raych1, line 154: log an error if swaggerFile is found from cache.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -111,6 +177,44 @@ export class TrafficValidator {
...
+        this.coverageResult.set(key, 0);
+        this.coverageData.push({
+          spec: key,
+          coveredOperaions: 0,
+          totalOperations: value.length,
+          coverageRate: 0,
+        });
+      } else {
+        if (value !== undefined && value.length !== 0) {
+          this.coverageResult.set(key, this.trafficOperation.get(key)!.length / value.length);
```
raych1: How about use a variable to keep `this.trafficOperation.get(key)` which is called triple times?

test/liveValidation/payloads/coveragetest/deleteCosmosTable_input.json
```diff
@@ -0,0 +1,35 @@
+{
```
raych1: Can you drop these payloads file into payloads/dataplane folder?

lib/swaggerValidator/trafficValidator.ts
raych1: Maybe I'm not clear with above message. There's a typo for `is` which should be `isn't`.  

What I meant for `cache` is `operationSpecMapper` instead of operation cache in live validator. So, pls revert the change to query from that cache.

lib/swaggerValidator/trafficValidator.ts
raych1: What's the change or comment for this issue?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -19,12 +29,24 @@ export interface RuntimeException {
   message: string;
 }
 
+export interface OperationCoverageInfo {
+  readonly spec: string;
+  readonly coveredOperaions: number;
```
raych1, line 34: Add validationFailOperations which count the operation if (validationResult.responseValidationResult.isSuccessful === false || undefined || validationResult.requestValidationResult.isSuccessful === false || undefined)

Add undefinedOperations to count unmatched traffic.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -105,7 +108,7 @@ export class TrafficValidator {
       try {
         spec = await this.loader.load(pathResolve(swaggerPath));
       } catch (e) {
-        console.log(e);
+        console.log(`Exception when loading spec ${e}`);
```
raych1: Replace ${e} by ErrorMessage:${e?.message};ErrorStack: ${e?.stack}.

lib/swaggerValidator/trafficValidator.ts
raych1: opInfo is removed from validationResult, so here we need to run operation search again.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -151,8 +156,25 @@ export class TrafficValidator {
...
           }
+          if (
+            validationResult.requestValidationResult.isSuccessful === false ||
+            validationResult.requestValidationResult.isSuccessful === undefined ||
+            validationResult.responseValidationResult.isSuccessful === false ||
+            validationResult.responseValidationResult.isSuccessful === undefined ||
+            validationResult.runtimeException !== undefined
+          ) {
+            if (this.validationFailOperations.get(swaggerFile) === undefined) {
+              this.validationFailOperations.set(swaggerFile, []);
```
raych1, line 169: Can you push operationId in this set and change followed if to else if?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -189,28 +211,38 @@ export class TrafficValidator {
...
-          spec: key,
-          coveredOperaions: 0,
-          totalOperations: value.length,
-          coverageRate: 0,
-        });
+        coveredOperaions = 0;
+        coverageRate = 0;
+        this.coverageData.set(key, 0);
       } else {
         if (value !== undefined && value.length !== 0) {
```
raych1: Can you use else if instead?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -44,9 +45,11 @@ export class TrafficValidator {
   private trafficPath: string;
   private loader?: LiveValidatorLoader;
   private trafficOperation: Map<string, string[]> = new Map<string, string[]>();
+  private validationFailOperations: Map<string, string[]> = new Map<string, string[]>();
+  private coverageData: Map<string, number> = new Map<string, number>();
   public operationSpecMapper: Map<string, string[]> = new Map<string, string[]>();
-  public coverageResult: Map<string, number> = new Map<string, number>();
-  public coverageData: OperationCoverageInfo[] = [];
+  public coverageResult: OperationCoverageInfo[] = [];
```
raych1: My bad, can you rename coverageResult to operationCoverageResult, rename undefinedResult to operationUndefinedResult  for better understanding?

lib/swaggerValidator/trafficValidator.ts
xiaoxuqi-ms: operaionId is pushed to this set already, do you mean to get all these operations together with the coverage result?

xiaoxuqi-ms: here we should use if but not else if: after the set is initiated, we need to push failed operationid.

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -108,7 +108,7 @@ export class TrafficValidator {
       try {
         spec = await this.loader.load(pathResolve(swaggerPath));
       } catch (e) {
-        console.log(`Exception when loading spec ${e}`);
+        console.log(`ErrorMessage: ${e?.message}; ErrorStack: ${e?.stack}.`);
```
raych1: Can you keep original extra message 'Exception when loading spec' either?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -219,24 +219,22 @@ export class TrafficValidator {
         coveredOperaions = 0;
         coverageRate = 0;
         this.coverageData.set(key, 0);
+      } else if (value !== undefined && value.length !== 0) {
+        coveredOperaions = this.trafficOperation.get(key)!.length;
+        coverageRate = coveredOperaions / value.length;
+        this.coverageData.set(key, coveredOperaions / value.length);
```
raych1: Use coverageRate variable.

raych1 (approved):

PR #759, opened 2022-02-24
Fix GC introduced by 2.11.7

xiaoxuqi-ms:


ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 02/24/2022 2.11.8
+
+- FIx GC introduced by 2.11.7
```
raych1, line 5: Can you change the log if the previous implementation doesn't cause GC?

lib/liveValidation/liveValidator.ts
```diff
@@ -803,6 +803,14 @@ export class LiveValidator {
     }
   }
 
+  private onlyUnique(value: string, index: number, self: string[]) {
+    return self.indexOf(value) === index;
+  }
+
+  private getUniqueArray(self: string[]) {
```
raych1: How about adding UT for this function?

ChangeLog.md
xiaoxuqi-ms: The GC is introduced by 2.11.7, this PR will fix this.

raych1 (approved):

PR #764, opened 2022-03-18
Xiaoxu/version: add version filter for legacy year

xiaoxuqi-ms:


raych1 (commented):
Can you put all the path compute logic in live validation service then leverage `this.options.swaggerPaths` to pass in the candidate paths for loading? It would be better to not couple this with live validator.

PR #767, opened 2022-03-25
fix bug about query parameter validation

ankhyk:
Ignore INVALID_TYPE validation in case of query parameter in string format


lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -192,6 +211,36 @@ export const ajvErrorToSchemaValidateIssue = (
   return result;
 };
 
+const shouldRevalidateError = (
```
raych1: How about have this function called as ReValidateIfNeed? Then put all the logic here returning modified errors array.

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -77,7 +78,25 @@ export class AjvSchemaValidator implements SchemaValidator {
       const result: SchemaValidateIssue[] = [];
       const isValid = validateSchema.validate.call(ctx, data);
       if (!isValid) {
-        ajvErrorListToSchemaValidateIssueList(validateSchema.validate.errors!, ctx, result);
+        const errors = validateSchema.validate.errors!;
+        const newErrors = lodash.cloneDeep(errors);
```
raych1: We don't want to impact all errors here and just move the re-validation logic to a signle method.

test/modelValidatorTests.ts
```diff
@@ -11,7 +11,7 @@ const testPath = __dirname;
 const specPath =
   `${testPath}/modelValidation/swaggers/specification/scenarios/resource-manager/` +
   `Microsoft.Test/2016-01-01/test.json`;
-
+jest.setTimeout(999999);
```
raych1: Will this need when run in pipeline or just need when run locally?

ChangeLog.md
```diff
@@ -1,6 +1,8 @@
 # Change Log - oav
 
 ## 03/11/2022 2.11.9
+
+- ModelValidator - Ignore INVALID_TYPE validation in case of query parameter in string format
```
raych1, line 5: You upgrade the package version to 2.11.10, changelog also needs to be aligned.

test/modelValidatorTests.ts
ankhyk: This is needed in pipeline。
Sometimes the test "should pass when specPath is a url starting with https://github/" will report a timeout error in pipeline.
Should I only change timeout of the test instead of set all timeout to "999999"?

lib/swaggerValidator/ajvSchemaValidator.ts
ankhyk: The error alone does not fully determine whether it can be skipped, query parameter array's item should also be revalidated

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -192,6 +201,49 @@ export const ajvErrorToSchemaValidateIssue = (
...
+      const arrayData = errorData.split(",").map((item) => {
+        if (parseFloat(item)) {
+          return parseFloat(item);
+        }
+        return item;
+      });
+      const position = dataPath.substr(1);
+      lodash.set(newData, position, arrayData);
+      const isValid = validate.call(ctx, newData);
+      if (isValid || !lodash.isEqual(validate.errors![i], error)) {
```
raych1: When will validate.errors![i] === error?

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -192,6 +201,49 @@ export const ajvErrorToSchemaValidateIssue = (
...
+    // If the value of query parameter is in string format, we can skip this error
+    if (
+      !ctx.isResponse &&
+      keyword === "type" &&
+      schema === "array" &&
+      typeof errorData === "string" &&
+      (parentSchema as any)?.["in"] === "query"
+    ) {
+      const arrayData = errorData.split(",").map((item) => {
+        if (parseFloat(item)) {
```
raych1: Can you catch exception when item is not number?

ankhyk: In addition to number and string, there is boolean that needs to be revalidated, revalidation has been added

lib/swaggerValidator/ajvSchemaValidator.ts
ankhyk: I have updated the logic of this part, which is used to add new errors to result

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -226,17 +226,41 @@ const ReValidateIfNeed = (
       (parentSchema as any)?.["in"] === "query"
     ) {
       const arrayData = errorData.split(",").map((item) => {
-        if (parseFloat(item)) {
+        // when item is number
+        const numberRegex = /^[+-]?\d+(\.\d+)?([Ee]\+?\d+)?$/g;
+        if (numberRegex.test(item)) {
+          // delete extra 0 in front and end of item
+          if (!item.includes("e") && !item.includes("E")) {
```
raych1: Remove this if not needed.

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -226,17 +226,41 @@ const ReValidateIfNeed = (
...
       const position = dataPath.substr(1);
       lodash.set(newData, position, arrayData);
       const isValid = validate.call(ctx, newData);
-      if (isValid || !lodash.isEqual(validate.errors![i], error)) {
-        continue;
+      if (!isValid) {
+        // if validate.errors have new errors, add them to result
+        validate.errors!.forEach((error) => {
+          let isNewError = true;
+          for (const error1 of errors) {
```
raych1: Can you rename errors to originalErrors?

test/modelValidation/swaggers/specification/query/examples/string.json
```diff
@@ -1,7 +1,6 @@
 {
   "parameters": {
-    "hello": "Wake up at 3:14:16 AM",
-    "helloArray": "50.97452,5.86605"
+    "hello": "Wake up at 3:14:16 AM"
```
raych1: Remove this example if this example is not used.

lib/swaggerValidator/ajvSchemaValidator.ts
```diff
@@ -226,17 +226,41 @@ const ReValidateIfNeed = (
...
+            if (lodash.isEqual(error, error1)) {
+              isNewError = false;
+            }
+          }
+          if (isNewError) {
+            result.push(validate.errors![i]);
```
raych1: Index i doesn't seem right here to get current error.

test/modelValidatorTests.ts
```diff
@@ -461,44 +461,40 @@ describe("Model Validation", () => {
...
+      });
+      it("should report other error and skip INVALID_TYPE error about query parameter", async () => {
+        const specPath2 = `${testPath}/modelValidation/swaggers/specification/query/test.json`;
+        const result = await validate.validateExamples(
+          specPath2,
+          "Query_StringButDefinedAsArray_extraError",
+          { consoleLogLevel: "off" }
+        );
+        assert(result.length === 2);
+        assert.strictEqual(result[0].code, "INVALID_TYPE");
```
raych1, line 491: Is this invalid_type error the same as previous error without re-validation?

raych1: Original error expects array type rather than string type.

raych1 (approved):

raych1 (approved):

PR #768, opened 2022-03-28
[ApiScenario] support readme

jordan13game:


lib/util/getAutorestConfigInternal.ts
```diff
@@ -46,7 +46,7 @@ const getAutorest = async (version: string | undefined, rfs?: IFileSystem, readm
 };
 
 const getAutorestCoreVersion = async (_args?: any) => {
-  return "latest-installed";
+  return "2.0.4421";
```
leniatgh: what's this for?

lib/commands/run-api-scenario.ts
```diff
@@ -122,6 +135,25 @@ export const builder: yargs.CommandBuilder = {
 
 export async function handler(argv: yargs.Arguments): Promise<void> {
   await cliSuppressExceptions(async () => {
+    const swaggerFilePaths: string[] = (argv.specs || []).map((it: string) => path.resolve(it));
+    if (argv.readme !== undefined) {
+      const readmeMd: string = path.resolve(argv.readme);
+      const autorestConfig = await getAutorestConfig(argv, readmeMd);
```
leniatgh: if `tag` is specified, we should load swagger files under it, otherwise, load the default tag.

lib/commands/generate-api-scenario.ts
```diff
@@ -2,81 +2,13 @@
...
-    console.log(`Transforming:\t${filePath}`);
-    const tracking = await recordingLoader.load(filePath);
-    trackingList.push(tracking);
-  }
-  await generator.generateTestDefinition(trackingList, output);
-
-  await generator.writeGeneratedFiles();
-  process.exit(0);
+export function builder() {
+  return yargs.commandDir("apiScenario");
```
leniatgh, line 11: pls update `documentation/api-scenario/how-to/generateABasicApiScenario.md` in spec repo accordingly.

lib/apiScenario/apiScenarioYamlLoader.ts
```diff
@@ -70,10 +71,17 @@ export class ApiScenarioYamlLoader implements Loader<RawScenarioDefinition> {
   }
 }
 
+const getAllSwaggerFilePathUnderDir = (dir: string): any[] => {
+  const allSwaggerFiles = fs
```
leniatgh, line 75: We should not use `fs`, assuming apiScenario file is in local fs. The apiScenario file could be a remote uri.

lib/util/getAutorestConfigInternal.ts
jordan13game: it can't run in autorest 3.x, so we need declare 2.x version

lib/commands/run-api-scenario.ts
jordan13game: Yes, it in the argv, and passed to autorest

lib/apiScenario/apiScenarioYamlLoader.ts
jordan13game: This is revert to previous version. So, how to fix it.

lib/commands/generate-api-scenario.ts
jordan13game: Sure

lib/apiScenario/apiScenarioYamlLoader.ts
leniatgh: there's `--specs` parameter in `run-api-scenario` command, so no need to scan swagger folders

lib/util/getAutorestConfigInternal.ts
leniatgh: 2.x has been deprecated. And why do we need autorest/core here?

lib/apiScenario/apiScenarioYamlLoader.ts
leniatgh: also, pls update the `apiScenarioSchema.ts` with the one in spec repo

lib/util/getAutorestConfigInternal.ts
jordan13game: using autorest core to resolve readme file. I have fixed to use 3.x version.

lib/apiScenario/apiScenarioYamlLoader.ts
jordan13game: this function is used to find swaggers when input swagger files is empty.  While this situation won't happen....

jordan13game: But we need to support previous version input, so leave it as what it was.

PR #769, opened 2022-03-30
Support traffic validation report generation

raych1:
Report including coverage data, error information.


lib/report/generateReport.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.language = language;
+    this.generalErrorResults = new Map();
+    this.setMetrics();
+    this.sortOperationIds();
+    this.prepareDataForRendering();
+  }
+
+  private prepareDataForRendering() {
+    const specLinkPrefix = "https://github.com/Azure/azure-rest-api-specs/blob/main/";
+    const payloadLinkPrefix = "https://github.com/scbedd/oav-traffic-converter/blob/main/sample-tables-input/"
```
ruowan: Why hard code this here? can this be a argument?

ruowan: Why hard code this here? can this be a argument?

lib/report/generateReport.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.generatedDate = new Date();
+    this.language = language;
+    this.generalErrorResults = new Map();
+    this.setMetrics();
+    this.sortOperationIds();
+    this.prepareDataForRendering();
+  }
+
+  private prepareDataForRendering() {
+    const specLinkPrefix = "https://github.com/Azure/azure-rest-api-specs/blob/main/";
```
xiaoxuqi-ms: Is it better to get these prefix from arguments?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -216,15 +230,27 @@ export class TrafficValidator {
...
     this.operationSpecMapper.forEach((value: string[], key: string) => {
+      unCoveredOperationsList = [];
       if (this.trafficOperation.get(key) === undefined) {
         coveredOperaions = 0;
         coverageRate = 0;
         this.coverageData.set(key, 0);
       } else if (value !== undefined && value.length !== 0) {
-        coveredOperaions = this.trafficOperation.get(key)!.length;
+        const validatedOperations = this.trafficOperation.get(key);
+        coveredOperaions = validatedOperations!.length;
```
xiaoxuqi-ms, line 243: use 'this.validatedOperations!.length;' ?

lib/swaggerValidator/trafficValidator.ts
```diff
@@ -216,15 +230,27 @@ export class TrafficValidator {
     let coveredOperaions: number;
     let coverageRate: number;
     let validationFailOperations: number;
+    let unCoveredOperationsList: OperationMeta[];
     this.operationSpecMapper.forEach((value: string[], key: string) => {
+      unCoveredOperationsList = [];
       if (this.trafficOperation.get(key) === undefined) {
```
xiaoxuqi-ms, line 237: if there is no operations found from traffic, should we also list the uncovered operations?

lib/report/generateReport.ts
raych1: Exposed this command line parameter.

raych1: Exposed this command line parameter.

ruowan (approved):

xiaoxuqi-ms (approved):

lib/swaggerValidator/trafficValidator.ts
xiaoxuqi-ms: For single RP validation, if there are no operations found from traffic for a certain swagger file, then the missing operations should be reported. But if it is multi-RP scenario, we can ignore this.

lib/swaggerValidator/trafficValidator.ts
raych1: there's no this.validatedOperations defined.

lib/swaggerValidator/trafficValidator.ts
raych1: When there's no operations found from traffic, uncovered operations is also reported. It works for both single RP and multiple RP.

PR #770, opened 2022-03-31
Generate Swagger Accuracy Report

JackTn:


lib/templates/baseLayout.mustache
```diff
@@ -1,319 +1,578 @@
...
+      }
+      a {
+        text-decoration: none;
+        position: relative;
+        display: inline-block;
+        color: var(--link-color);
+      }
+      a span {
+        margin: auto;
+        display: inline-block;
```
raych1: I tried to render and needs to remove {display:.; position:.; background:.} style of `a span`. Can you check?

raych1 (approved):

PR #776, opened 2022-04-06
Fix bug about error path

ankhyk:


lib/util/jsonUtils.ts
```diff
@@ -166,7 +166,11 @@ export const getFilePositionFromJsonPath = (
   const pathArr = jsonPathToArray(jsonPath.substr(1));
   const newPathArr = pathArr.slice(0);
   const index = newPathArr.findIndex((str) => str.includes("/providers/Microsoft"));
-  if (index !== -1 && newPathArr[index + 1] !== undefined) {
```
raych1, line 169: can you add comment about the error path to be processed?

test/semanticValidatorTests.ts
```diff
@@ -62,6 +62,14 @@ describe("Semantic validation", () => {
         constants.ErrorCodes.InternalError.name
       );
     });
+    // error path
+    it("should report truly error when error path contains '/providers/Microsoft'", async () => {
+      const specPath = `${testPath}/semanticValidation/specification/validateSwaggerSchema/internalErrors/errorPath.json`;
+      const result = await validate.validateSpec(specPath, undefined);
+      assert(result.validityStatus === false);
+      assert.strictEqual(result.validateSpec?.errors?.[0].code, "ENUM_MISMATCH");
```
raych1, line 70: As this test is to validate the error path, can you add the assert on the error path?

lib/util/jsonUtils.ts
raych1: can you update comment as: when jsonPath='/providers/Microsoft.Provider/resource', the split pathArr will be ['/providers/Microsoft','Provider/resource'].  Only in this case, these two elements in the array need to be composed together by '.'. So restrict the condition to the path element ends with /providers/Microsoft.

raych1 (approved):

PR #777, opened 2022-04-06
Update azure-pipelines.yml for Azure Pipelines

ruowan:


raych1:
Closed this PR as the error is caused by one package reference to ADO registry.

PR #779, opened 2022-04-13
Fix CG

ankhyk:
- Bump json-pointer from 0.6.1 to 0.6.2
- Bump async from 3.2.0 to 3.2.3
- Bump ansi-regex to 5.0.1
- Bump md5-file from 4.0.0 to 5.0.0
- Bump validator to 13.7.0
- Bump json-schema from 0.2.3 to 0.4.0

list of dependencies in CG alerts
- [ ] async ≥ 3.2.2 ( yasway => istanbul ) ( can’t fix )
- [x] ansi-regex ≥ 4.1.1 ( already fixed )
- [ ] qs ≥ 6.8.1 ( request, postman-request) ( can’t fix )
- [x] validator ≥ 13.7.0
- [x] json-pointer ⇒ latest
- [x] md5-file ⇒ latest
- [x] json-schema ≥ 0.4.0


package.json
```diff
@@ -60,8 +60,7 @@
     "winston": "^3.2.1",
     "yargs": "^15.4.1",
     "yasway": "^1.10.7",
-    "yuml2svg": "^4.2.1",
-    "z-schema": "^4.2.2"
```
raych1, line 64: Those references to types in z-schema needs to be refactored too.

ankhyk: This line is just remove extra "z-schema" dependency because "swagger-parser" and "yasway" both have "z-schema" dependency.

Security alert said it can't update "validator" to 13.7.0 because "validator" is under "z-schema", here's the list about "validator"
```
+-- swagger-parser@6.0.5
| `-- z-schema@3.25.1
|   `-- validator@10.11.0
`-- yasway@1.10.7
  `-- z-schema@4.2.3
    `-- validator@12.2.0
```
To update "validator" to 13.7.0, we can update "swagger-parser" to 10.0.3. But the latest version of "[yasway](https://www.npmjs.com/package/yasway)" is 1.10.7, so the alert will still be there.

package-lock.json
```diff
@@ -30456,6 +29909,17 @@
           "requires": {
             "isarray": "0.0.1"
           }
+        },
+        "z-schema": {
+          "version": "4.2.4",
```
raych1, line 29914: Should we change this to 5.0.3? and validator:"^13.7.0"?

ankhyk: No, "z-schema" v4.2.4 can use "validator" ≥ 13.6.0
```
> npm list validator

├─┬ yasway@1.10.7
│ └─┬ z-schema@4.2.4
│   └── validator@13.7.0 deduped
└─┬ z-schema@5.0.3
  └── validator@13.7.0
```

raych1 (approved):

PR #784, opened 2022-04-25
Support remote uri

leniatgh:


lib/swagger/fileLoader.ts
```diff
@@ -46,22 +47,24 @@ export class FileLoader implements Loader<string> {
     if (this.opts.fileRoot) {
       filePath = this.resolvePath(filePath);
       const url = urlParse(filePath);
-      if (url === undefined) {
-        filePath = pathRelative(this.opts.fileRoot, filePath);
-      } else {
+      if (url) {
         const rootUrl = urlParse(this.opts.fileRoot);
-        filePath = rootUrl === undefined ? filePath : pathRelative(rootUrl.path, url.path);
+        filePath = rootUrl === undefined ? filePath : path.relative(rootUrl.path, url.path);
```
jordan13game, line 52: @azure-tools/openapi-tools-common pathRelative?

leniatgh: there's no pathRelative inside it. The implementation is same. So using path.relative instead.

PR #795, opened 2022-05-06
fix example based template generate

jordan13game:


lib/apiScenario/apiScenarioLoader.ts
```diff
@@ -510,6 +509,18 @@ export class ApiScenarioLoader implements Loader<ScenarioDefinition> {
       step.responseExpected = exampleFileContent.responses;
 
       await this.applyPatches(step, rawStep);
+      const getVariable = (name: string) => {
+        const variable =
+          step.variables[name] ?? ctx.scenario?.variables[name] ?? ctx.scenarioDef.variables[name];
+        if (variable === undefined) {
+          const requiredVariables =
+            ctx.scenario?.requiredVariables ?? ctx.scenarioDef.requiredVariables;
```
leniatgh, line 517: what if `scenario.requireVariables` is not empty, but doesn't include `name` ?

leniatgh (commented):
could you also update the version?

lib/apiScenario/apiScenarioLoader.ts
jordan13game: return ```undefined```, templateGenerator only check whether it is ```undefined```. If not, it will replace with ```$(${name})```.

PR #797, opened 2022-05-12
Generate swagger report 2

JackTn:


lib/commands/validate-traffic.ts
```diff
@@ -43,6 +43,12 @@ export const builder: yargs.CommandBuilder = {
     boolean: true,
     default: false,
   },
+  runtimeException: {
```
raych1: Can you rename `runtimeException` to `outputExceptionInReport`, alias to "oe"? 

describe: "Whether rendering the runtime exceptions in the validation report"

raych1 (approved):

Issue #811, opened 2022-06-06
Support validating duration format


leniatgh:
The value `"disconnectDelay": "00:05",` is not valid `duration` format, defined here: https://github.com/Azure/azure-rest-api-specs/pull/16557/files#diff-3c4db78d53e5371a8535d5e3605c358012d88939dacc237a6a8441b4bbed68caR257

`duration` is new format added in Draft 2019-09, since we allow specifying in swagger, we should validate the example value as well.

raych1:
@keryul , I remember this is supported. Can you check?

raych1:
This format is supported in oav@3.0.1

Issue #805, opened 2022-06-01
Multipart-formdata not supported with MULTIPLE_BODY_PARAMETERS


alexhcheng:
following the spec here: https://swagger.io/docs/specification/2-0/file-upload/

snippet of swagger
```
   "/MyMethod": {
      "post": {
        "summary": "test",
        "description": "test",
        "operationId": "MyOperationId",
        "consumes": [
          "multipart/form-data"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "in": "query",
            "name": "q1",
            "description": "test",
            "type": "string"
          },
          {
            "in": "query",
            "name": "q2",
            "description": "test",
            "required": true,
            "type": "string"
          },
          {
            "in": "formData",
            "name": "Document",
            "description": "Actual Document in the form",
            "required": true,
            "type": "file"
          },
          {
            "in": "formData",
            "name": "Glossary",
            "description": "Glossary for the request",
            "type": "file"
          }
        ],
```

Running the validation gives MULTIPLE_BODY_PARAMETERS error

alexhcheng:
swagger renders properly in swagger UI

raych1:
This is known issue and we plan to fix it in this sprint.

alexhcheng:
Just wondering when the release date will be so i can report back to my team.  thanks! @raych1

raych1:
@alexhcheng , it's planned to release by end of this month. One thing want to confirm with you, your swagger example above has two formData parameters with `file` type. It's not supported from the openapi document https://swagger.io/docs/specification/2-0/file-upload/. Quote of `However, uploading an arbitrary number of files (an array of files) is not supported`. 

For multiple formData parameters, only one parameter can be `file` type.

alexhcheng:
Hi @raych1, arbitrary number of files are not supported.  But fixed number of files are.  According to the spec.  In the link you sent, under the "Multiple Upload" Section, it states: "You can have several named file parameters, each defined individually".  

This has multiple "file" type, each with different "name".  This format needs to be supported.
```
      parameters:
        - in: formData
          name: upfile1
          type: file
          required: true
        - in: formData
          name: upfile2
          type: file
          required: false
        - in: formData
          name: upfile3
          type: file
          required: false
```

ankhspirit:
@alexhcheng , This issue was resolved in version 3.0.1

ankhyk:
@alexhcheng , This issue was resolved in version 3.0.1

Issue #791, opened 2022-05-05
Generated examples use non-standard example domain


abatishchev:
Currently they're generating this:

```json
{
"headers": {
    "Azure-AsyncOperation": "https://foo.com/operationstatus"
  }
}
```

However there are better options:

- [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606) suggests to use `example.com`
- [Microsoft REST API Guideline](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md)  uses `contoso.com`
- Use `management.azure.com`, always or figure it out based on some input or somehow else

raych1:
This request is resolved in oav@3.0.1

Issue #786, opened 2022-05-04
oav doesn't understand format: duration, generates string which fails validation


abatishchev:
Cadl:

```
@doc("Category properties")
model CategoryProperties {
  @doc("Window duration")
  windowDuration: duration;
}
```

Swagger:

```json
"CategoryProperties": {
  "type": "object",
  "properties": {
    "windowDuration": {
      "type": "string",
      "format": "duration",
      "description": "Window duration"
    }
  },
  "description": "Category properties",
  "required": [
    "windowDuration"
  ]
}
```

Example:

```json
{
  "parameters": {
    "api-version": "2022-06-01-preview",
    "subscriptionId": "aaaaaaaaaaaaaaaaaaaa",
    "resourceGroupName": "rgthrottling",
    "accountName": "aaaaa",
    "categoryName": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
    "fixedThresholdName": "aaaaaaaaaaaaaaaaaaaaaa",
    "properties": {
      "properties": {
        "value": 24,
        "depth": 16,
        "minDecisionDuration": "aaaaaaaaaaaaaaaaaaaaaaaaa"
      }
    }
  },
  "responses": {
    "200": {
      "body": {
        "properties": {
          "value": 14,
          "depth": 10,
          "minDecisionDuration": "aaaaaaaaaaaaaaa"
        },
        "systemData": {
          "createdBy": "a",
          "createdByType": "User",
          "createdAt": "2022-05-04T20:28:44.075Z",
          "lastModifiedBy": "a",
          "lastModifiedByType": "User",
          "lastModifiedAt": "2022-05-04T20:28:44.075Z"
        },
        "name": "aaaaaaaaaaaaa",
        "id": "aaaaaaaaaa",
        "type": "aaaaaaaaaaaaaa"
      }
    }
  }
}
```

Validation error:

```
{
  code: 'INVALID_FORMAT',
  message: "Object didn't pass validation for format duration: aaaaaaaaaaaaaaa",
  schemaUrl: 'specification/throttling/resource-manager/Microsoft.Throttling/preview/2022-06-01-preview/throttling.json',
  exampleUrl: 'specification/throttling/resource-manager/Microsoft.Throttling/preview/2022-06-01-preview/examples/FixedThresholds_Update_MaximumSet_Gen.json',
  schemaPosition: { line: 1310, column: 32 },
  schemaJsonPath: '#/properties/minDecisionDuration/format',
[...]

PR #812, opened 2022-06-09
Fix issues about: multipart/form-data, duration, exampleJsonPath, LRO validation and generate

ankhyk:
- GenerateExamples - Change the domain of LRO header
- ModelValidator - Still validate LRO operation when response doesn't have schema
- ModelValidator - Report real exampleJsonPath when additional parameter includes "."
- Support duration format when generating examples
- SemanticValidator - Fix bug about multipart/form-data

#805 #811 #791 #786


lib/swaggerValidator/semanticValidator.ts
```diff
@@ -509,13 +510,24 @@ export class SwaggerSemanticValidator {
...
+              NumberOfFileTypeFormData++;
+            }
             if (bodyParam !== undefined) {
               const meta = getOavErrorMeta(
                 param.in === bodyParam.in
                   ? "MULTIPLE_BODY_PARAMETERS"
                   : "INVALID_PARAMETER_COMBINATION",
                 {}
               );
+              if (
```
raych1, line 519: How about the case that have multiple formData parameters but without body parameter?

raych1: parameters sequence isn't fixed, body param might between multiple form data parameters. So don't need to check on the meta.code. And also assign the bodyParam before continue.

test/modelValidatorTests.ts
```diff
@@ -110,6 +110,15 @@ describe("Model Validation", () => {
         assert.strictEqual(err.innerErrors[0].code, "DOUBLE_FORWARD_SLASHES_IN_URL");
       }
     });
+
+    it("should report real exampleJsonPath when additional parameter includes '.'", async () => {
+      const specPath2 = `${testPath}/modelValidation/swaggers/specification/additionalParameter/additionalParameter.json`;
```
raych1, line 115: Is this test for exampleJson path change?

test/__snapshots__/exampleGeneratorTests.ts.snap
```diff
@@ -125,139 +307,4344 @@ Array [
...
-    "examplePosition": Object {
-      "column": 16,
-      "line": 16,
-    },
-    "exampleUrl": "test/exampleGenerator/specification/sql/resource-manager/Microsoft.Sql/preview/2020-02-02-preview/examples/ManagedDatabases_CreateOrUpdate_MaximumSet_Gen.json",
-    "message": "Response statusCode 201 for operation ManagedDatabases_CreateOrUpdate has no response body provided in the example, however the response does have a \\"schema\\" defined in the swagger spec.",
-    "operationId": "ManagedDatabases_CreateOrUpdate",
+    "code": "LRO_RESPONSE_HEADER",
+    "exampleJsonPath": "responses/202/headers",
+    "examplePosition": undefined,
```
raych1, line 313: This is regression, position should not be undefined.

test/modelValidatorTests.ts
ankhyk: yes, https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1505953/, this is the work item

regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -10122,30 +11257,62 @@ Array [
...
+      "column": 22,
+      "line": 174,
     },
     "schemaUrl": "/home/vsts/work/1/s/regression/azure-rest-api-specs/specification/iotcentral/resource-manager/Microsoft.IoTCentral/stable/2018-09-01/iotcentral.json",
     "severity": 0,
-    "source": "request",
+    "source": "response",
   },
   Object {
-    "code": "PATTERN",
```
raych1, line 10144: Why this error is gone?

raych1 (approved):

PR #827, opened 2022-06-29
Save payload

leniatgh:


jordan13game (approved):

lib/apiScenario/gen/staticApiScenarioGenerator.ts
```diff
@@ -52,13 +38,13 @@ export class StaticApiScenarioGenerator {
       eraseDescription: false,
       filerTopLevelResourceType: true,
       noExternalDependencyResourceType: true,
-      rules: ["resource-put-delete"],
```
raych1, line 55: Is this rule used currently? Is it safe to delete? @ruowan

raych1:
@ruowan , there're a bunch of deletions, can you review those changes?

lib/apiScenario/gen/staticApiScenarioGenerator.ts
ruowan: Currently, we don't use this. Unified pipeline use fixed version of oav. It's safe to delete it.

ruowan:
LGTM. For dependency we rely on restler and report has been already refactor as readme or html format.

ruowan (approved):

PR #834, opened 2022-07-11
Support data validation and generation in byte format

ankhyk:


tsconfig.json
```diff
@@ -55,7 +55,7 @@
     /* Experimental Options */
     "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-    "lib": ["es2020.string"],
+    "lib": ["es2020.string", "dom"],       /* Enables function atob(). */
```
raych1: I don't see any reference to atob function.

ankhyk: canceled, I used atob() at first, but finally switched to Buffer.from()

raych1 (approved):

PR #845, opened 2022-08-02
Add type validation for additionalProperties

ankhyk:


regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
```diff
@@ -5653,6 +5653,462 @@ Array [
...
   },
+  Object {
+    "code": "INVALID_TYPE",
+    "exampleJsonPath": "$responses.200.body.value[0].properties.activities[0].typeProperties.activities[0]['inputs']",
+    "examplePosition": Object {
+      "column": 35,
+      "line": 46,
+    },
+    "exampleUrl": "/home/vsts/work/1/s/regression/azure-rest-api-specs/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/examples/Pipelines_ListByFactory.json",
+    "message": "Expected type object but found type array",
```
raych1: Can you check if these new errors are expected and caused by this PR change?

ankhyk: some errors are unexpected, I'll update the rules

lib/swagger/swaggerTypes.ts
```diff
@@ -268,6 +268,9 @@ export interface Schema extends BaseSchema {
 
   // x-ms-discriminator-value exists but discriminator missing
   _missingDiscriminator?: boolean;
+
+  // Additional property support
+  additionalObjectTypeProperties?: boolean;
```
raych1: Can you rename it to `additionalPropertiesWithObjectType`

regression/__snapshots__/validateExamplesRegression_latestOnly.test.ts.snap
raych1: Did you make the change? I don't see updates to this PR.

ankhyk:
I validated the five snapshot test and the result was as expected

raych1 (approved):

PR #853, opened 2022-08-18
Add markdown to swagger report  

JackTn:


lib/report/generateReport.ts
```diff
@@ -273,6 +278,12 @@ export class CoverageView {
     }
   }
 
+  private async readMarkdown() {
+    const loader = new FileLoader({});
```
raych1: can you add try catch here so that markdown loading failure doesn't block html report rendering?

raych1 (changes requested):
Others look good.

raych1 (approved):

PR #854, opened 2022-08-18
Support api-version validation

ankhyk:


lib/swaggerValidator/modelValidator.ts
```diff
@@ -390,6 +390,22 @@ export class SwaggerExampleValidator {
...
+          });
+          this.addErrorsFromErrorCode(
+            operation.operationId!,
+            exampleFileUrl,
+            meta,
+            operation,
+            undefined,
+            p
+          );
+          break;
```
raych1: continue instead of break here as we want to report all errors.

raych1 (approved):

raych1 (changes requested):
can you sample check the errors updated in snapshot ensure there's no regression?

In addition, can you fill in the value of exampleJsonPath and exampleJsonLocation in error details?

raych1 (approved):

PR #855, opened 2022-08-19
Fix bug when operation has no own parameters

ankhyk:


lib/swaggerValidator/modelValidator.ts
```diff
@@ -420,7 +420,7 @@ export class SwaggerExampleValidator {
       ajvValidatorErrors,
       exampleFileUrl,
       exampleContent,
-      parameters
+      parameters !== undefined ? parameters : publicParameters
```
raych1: Should we use mergedParameters instead?

ankhyk: Sure

raych1 (approved):

PR #862, opened 2022-09-05
support example used in multi operations

jordan13game:


lib/apiScenario/apiScenarioLoader.ts
```diff
@@ -427,7 +427,66 @@ export class ApiScenarioLoader implements Loader<ScenarioDefinition> {
...
+        if (operation === undefined) {
+          throw new Error(`Operation not found for ${step.operationId} in step ${step.step}`);
+        }
+        if (this.opts.includeOperation) {
+          step.operation = operation;
+        }
+      } else {
+        const opMap = this.exampleToOperation.get(exampleFilePath);
+        if (opMap === undefined) {
+          throw new Error(`Example file is not referenced by any operation: ${step.exampleFile}`);
```
leniatgh: we should use `rawStep.operationId` first, and then try to find operation from exampleFile

lib/apiScenario/apiScenarioSchema.ts
```diff
@@ -384,6 +384,9 @@ export const ApiScenarioDefinition: Schema & {
           type: "string",
           format: "uri-reference",
         },
+        operationId: {
+          type: "string",
+        },
```
leniatgh, line 389: please submit PR to update schema.json in spec repo too

leniatgh (changes requested):

leniatgh (approved):

PR #864, opened 2022-09-06
[Report] simulate hash click by jQuery

JackTn:


lib/templates/baseLayout.mustache
```diff
@@ -741,6 +750,25 @@
             });
         });
     }
+    
+    function hashLinkClick() {
+        $('span.hashLinkClass').each(function (index) {
+           $(this).click(function(event){
+                var disableCon = $(this).parent().parent().parent().parent().parent().parent().parent().parent().next()
```
raych1, line 757: Can we search for specified element or check on element attribute instead of using a bunch of parent()?

JackTn: Why we should use this parent().parent() ? This code is binding the element we will click and It will trigger a specific record of this report. If using specified element, it is unique and cannot support click in multi swagger report. On the other hand, if using specified element when get element, it will shows more element and we can not know which part we clicked!

raych1 (approved):

PR #865, opened 2022-09-07
remove circular ref

xiaoxuqi-ms:


lib/armValidator/operationLoader.ts
```diff
@@ -15,6 +15,12 @@ export enum CompareType {
   isMissing,
 }
 
+const statusCodeMap = new Map([
+  ["OK", "200"],
+  ["CREATED", "201"],
+  ["ACCEPTED", "202"],
```
raych1, line 21: Should we add 204?

lib/swagger/jsonLoader.ts
```diff
@@ -148,6 +148,9 @@ export class JsonLoader implements Loader<Json> {
             },
           },
         },
+        dereference: {
+          circular: "ignore",
```
raych1, line 152: Is this covered in test?

lib/armValidator/operationLoader.ts
xiaoxuqi-ms: Good catch!

lib/swagger/jsonLoader.ts
xiaoxuqi-ms: Add a test for circular swagger

raych1 (approved):

PR #869, opened 2022-09-26
replace variable in key

### LINKED ISSUE #844: 'subscriptionId' and 'resourceGroupName' in object key are not replaced
image

jordan13game:
resolves https://github.com/Azure/oav/issues/844


test/apiScenario/fixtures/specification/storage/resource-manager/Microsoft.Storage/stable/2021-08-01/scenarios/testKeyReplace.yaml
```diff
@@ -0,0 +1,11 @@
...
+scope: ResourceGroup
+
+scenarios:
+  - scenario: StorageBasicExample
+    steps:
+      - step: createStorageAccount
+        exampleFile: ../examples/StorageAccountCreateUserAssignedIdentityWithFederatedIdentityClientId.json
+        requestUpdate:
+          - move: /parameters/identity/userAssignedIdentities/~1subscriptions~1$(subscriptionId)~1resourceGroups~1$(resourceGroupName)~1providers~1Microsoft.ManagedIdentity~1userAssignedIdentities~1{managed-identity-name}
+            from: /parameters/identity/userAssignedIdentities/~1subscriptions~1{subscription-id}~1resourceGroups~1res9101~1providers~1Microsoft.ManagedIdentity~1userAssignedIdentities~1{managed-identity-name}
```
leniatgh: can we add one more test case using `replace`?

lib/apiScenario/gen/restlerApiScenarioGenerator.ts
```diff
@@ -411,7 +411,8 @@ export class RestlerApiScenarioGenerator {
         continue;
       }
       for (const method of Object.keys(dependencies[path])) {
-        const operationId = this.getOperationId(path, method);
+        const operationId =
+          this.getOperationId(path, method) || this.getOperationId(path + "/", method);
```
leniatgh, line 416: what's this for?

lib/apiScenario/newmanReportValidator.ts
```diff
@@ -205,6 +205,9 @@ export class NewmanReportValidator {
         }
 
         it.assertions.forEach((assertion) => {
+          if (assertion.test.includes("response status code assertion")) {
```
leniatgh: we can define a constant string in PostmanHelper, for better maintainability

leniatgh: It seems some assertions are implemented under the response status code assertion, we should take those out

leniatgh (commented):
Looks good, some minor comments

lib/apiScenario/gen/restlerApiScenarioGenerator.ts
jordan13game: some paths end with "/", but RESTler will remove it in dependencies. So we can't find it in swagger

leniatgh (approved):

PR #878, opened 2022-10-10
Fix tests on MacOS (case-insensitive fs)

### LINKED ISSUE #875: oav tests are failing in develop
There are tests failing on the develop branch.
```
> npm install
> npm run build
> npm test

...

Snapshot Summary
 › 26 snapshots failed from 2 test suites. Inspect your code changes or run `npm test -- -u` to update them.

Test Suites: 2 failed, 28 passed, 30 total
Tests:       63 failed, 4 skipped, 198 passed, 265 total
Snapshots:   26 failed, 29 passed, 55 total
Time:        33.509 s, estimated 38 s
Ran all test suites.
```

mikekistler:
This PR fixes test failures that occur when running the tests on MacOS or any system configured with a case-insensitive filesystem.

The test failures occur because the "getSuppression" method uses the `isUnderFileRoot` to check if the readmePath is under to root path for the file loader, but this method gives an incorrect return value (true) when the readmePath is an absolute path (as it is in this case). The file load then attempts to open readmePath which fails because the open method _correctly_ determines that the file is not under the file root.

The fix is to simply wrap the file open in a try catch that returns an empty array when the file is not under the file root.

I also added a bare-bones CONTRIBUTING.md since this would have been helpful to me when debugging this problem and developing the fix.

Fixes #875.


lib/swagger/suppressionLoader.ts
```diff
@@ -65,10 +65,16 @@ export class SuppressionLoader implements Loader<void, SwaggerSpec> {
   }
 
   private async getSuppression(readmePath: string): Promise<SuppressionItem[]> {
-    if (!this.fileLoader.isUnderFileRoot(readmePath)) {
-      return [];
+    // FileLoader.isUnderFileRoot() is broken for absolute paths, so we'll just catch and suppress the error
```
raych1, line 68: We will fix `isUnderFileRoot` function if it has a bug rather than bypass it.

raych1 (commented):
Thanks @mikekistler for making this PR. It's really helpful to have contributing document.

lib/swagger/suppressionLoader.ts
mikekistler: @raych1 What is the status on the fix to `isUnderFileRoot`?

raych1:
The fix was merged in another PR.

PR #879, opened 2022-10-11
Xiaoxu/addinfo

xiaoxuqi-ms:


lib/armValidator/roundTripValidator.ts
```diff
@@ -106,22 +106,35 @@ export function diffRequestResponse(
...
+  path: string,
+  it: any
+): LiveValidationIssue {
   let severity, message;
   switch (errorCode) {
     case "ROUNDTRIP_INCONSISTENT_PROPERTY": {
+      const details = ` Value in request: "${it.oldValue}", Value in response: "${it.value}", path: "${path}".`;
       severity = roundTripValidationErrors.ROUNDTRIP_INCONSISTENT_PROPERTY.severity;
-      message = roundTripValidationErrors.ROUNDTRIP_INCONSISTENT_PROPERTY.message({});
+      message = roundTripValidationErrors.ROUNDTRIP_INCONSISTENT_PROPERTY.message({}) + details;
```
leniatgh: what's the `{}` for? Can we put the value inside the message? like:
`The property's value 'westcentralus1' in the GET response is different from what was set 'westcentralus' in the preceding PUT request.`

test/__snapshots__/roundtripValidatorTests.ts.snap
```diff
@@ -86,7 +86,7 @@ Object {
     Object {
       "code": "ROUNDTRIP_ADDITIONAL_PROPERTY",
       "jsonPathsInPayload": Array [],
-      "message": "The property is returned in the GET response, but it is not declared in the PUT request.",
+      "message": "The property is returned in the GET response, but it is not declared in the PUT request. Additional value is \\"ReadWrite\\", path: \\"/properties/storageProfile/osDisk/caching\\".",
```
leniatgh: just realized the path is already there, so it seems we can skip outputting path in message, just output the property name (`caching` in this case)

lib/armValidator/roundTripValidator.ts
```diff
@@ -106,22 +106,35 @@ export function diffRequestResponse(
...
       break;
     }
     case "ROUNDTRIP_ADDITIONAL_PROPERTY": {
+      let details = " Type of additional value is object.";
+      if (it.value !== Object(it.value)) {
```
leniatgh: we can use `instanceof Object`

leniatgh (changes requested):

test/__snapshots__/roundtripValidatorTests.ts.snap
leniatgh: `The property 'caching' is returned in the GET response, but it is not declared in the PUT request.`

leniatgh (approved):

test/__snapshots__/roundtripValidatorTests.ts.snap
```diff
@@ -69,7 +69,7 @@ Object {
     Object {
       "code": "ROUNDTRIP_MISSING_PROPERTY",
       "jsonPathsInPayload": Array [],
-      "message": "The property is present in the PUT request but is either never returned in the GET response or is returned with a null value.",
+      "message": "The property \\"createOption\\" is present in the PUT request but is either never returned in the GET response or is returned with a null value.",
```
leniatgh: can we use single quote here? so no need to escape

Issue #84, opened 2017-04-05
Export two methods: “init” to build lookup and “validate” to validate request


vishrutshah:
- [x] CIT – write unit tests for the package
- [x] “config” - Enable consumption of configuration into the package for features like (rules on/off, github repo to do init on, etc…)

vishrutshah:
completed

PR #881, opened 2022-10-12
fix bug on isUnderFileRoot

JackTn:


lib/swagger/fileLoader.ts
```diff
@@ -79,11 +79,14 @@ export class FileLoader implements Loader<string> {
   }
 
   public isUnderFileRoot(filePath: string) {
-    if (!this.opts.fileRoot) {
+    if (this.opts.fileRoot === undefined || this.opts.fileRoot === null) {
+      return false;
+    } else if (!this.opts.fileRoot) {
```
raych1: Line#84 and line #85 aren't needed any more.

raych1:
@mikekistler , can you get this PR branch and validate if it's fixed in macos? Shanghai team cannot find a mac :-(

mikekistler:
This branch does not fix the problem. I created a test that should help you verify the fix on non-Mac systems:

https://github.com/mikekistler/oav/blob/file-loader-test/test/swagger/fileLoaderTests.ts

mikekistler:
What's the status on this fix?

JackTn:
@mikekistler  can you try this test for this branch again ? Thanks !

test/swagger/fileLoaderTest.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import assert from "assert";
+import { FileLoader, FileLoaderOption } from "../../lib/swagger/fileLoader";
+
+describe("FileLoader functions", () => {
+  describe("isFileUnderRoot", () => {
```
mikekistler:
```suggestion
  describe("isUnderFileRoot", () => {
```

lib/swagger/fileLoader.ts
```diff
@@ -79,11 +79,15 @@ export class FileLoader implements Loader<string> {
   }
 
   public isUnderFileRoot(filePath: string) {
-    if (this.opts.fileRoot === undefined || this.opts.fileRoot === null) {
-      return false;
+    if (!this.opts.fileRoot) {
```
raych1, line 82: I think line#82-#84 is not needed. No matter opts.fileRoot is provided or not, we need to check if the path is under root.

JackTn: if this.opts.fileRoot's type is string | undefined. If fileRoot is undefined, the path is under root or not?

mikekistler (approved):
These changes work on Mac. Thanks.

lib/swagger/fileLoader.ts
JackTn: one question. In origin code, if opts.fileRoot is undefined, it return true. But now, it always retrun false when opts.fileRoot is undefined. Is it correct ? @raych1

raych1: Let's keep original behavior.

JackTn: ok and it can be merged now @raych1

raych1 (approved):

PR #883, opened 2022-10-18
loosen validation for duration check

tadelesh:
For some swagger [examples](https://github.com/Azure/azure-rest-api-specs/blob/ae227e20bcb22d83634452026f701d75bec2619e/specification/monitor/resource-manager/Microsoft.Insights/stable/2018-03-01/examples/createOrUpdateMetricAlertSingleResource.json#L17), durations are not case sensitive. Loosen validation regex could make them compatible. But if it should be tightly aligned with spec, just correct me.


lib/swaggerValidator/ajv.ts
```diff
@@ -116,7 +116,7 @@ export const ajvEnableDurationFormat = (ajv: Ajv) => {
   ajv.addFormat("duration", {
     type: "string",
     validate:
-      /^P([0-9]+(?:[,\.][0-9]+)?Y)?([0-9]+(?:[,\.][0-9]+)?M)?([0-9]+(?:[,\.][0-9]+)?D)?(?:T([0-9]+(?:[,\.][0-9]+)?H)?([0-9]+(?:[,\.][0-9]+)?M)?([0-9]+(?:[,\.][0-9]+)?S)?)?$/,
+      /^P([0-9]+(?:[,\.][0-9]+)?Y)?([0-9]+(?:[,\.][0-9]+)?M)?([0-9]+(?:[,\.][0-9]+)?D)?(?:T([0-9]+(?:[,\.][0-9]+)?H)?([0-9]+(?:[,\.][0-9]+)?M)?([0-9]+(?:[,\.][0-9]+)?S)?)?$/i,
```
raych1, line 119: The character format must be uppercase from durations definition at https://en.wikipedia.org/wiki/ISO_8601

raych1 (changes requested):
@tadelesh,  it doesn't recognize the lowercase in durations format.

tadelesh:
> @tadelesh, it doesn't recognize the lowercase in durations format.

Some libraries loosen the condition to lowercase. But if you think it should be fixed with example, I'm OK with that. Besides, will lint rule check the example value of duration?

raych1:
> > @tadelesh, it doesn't recognize the lowercase in durations format.
> 
> Some libraries loosen the condition to lowercase. But if you think it should be fixed with example, I'm OK with that. Besides, will lint rule check the example value of duration?

No, it just validates the format instead of the value correctness.

PR #889, opened 2022-10-26
support data plane recording generate

jordan13game:


lib/apiScenario/gen/testRecordingApiScenarioGenerator.ts
```diff
@@ -168,7 +171,7 @@ export class TestRecordingApiScenarioGenerator {
   ): Promise<RawScenario> {
     logger.info(`\nGenerating ${requestTracking.description}`);
     const testScenario: RawScenario = {
-      scenario: requestTracking.description,
+      scenario: requestTracking.description.replace(/\(/g, "_").replace(/\)/g, "_"),
```
leniatgh:
```suggestion
      scenario: requestTracking.description.replace(/[\(\)]/g, "_"),
```
And any other special characters? like space,`!@#%^&*`?

lib/apiScenario/variableEnv.ts
```diff
@@ -262,6 +262,31 @@ export class VariableEnv {
     }
   }
 
+  public tryResolveString(source: string): string {
+    const regex = variableRegex;
```
leniatgh: directly use `variableRegex` ?

leniatgh: what's the difference with `doResolveStringVariable`?

leniatgh (changes requested):

lib/apiScenario/gen/testRecordingApiScenarioGenerator.ts
jordan13game: space may in filename, `!@#%^&*` is illegal in path

lib/apiScenario/variableEnv.ts
jordan13game: This function uses `get` not `getRequired`, it won't throw error when it not found. And only resolve `string` type without `bool` and `int`

leniatgh (approved):
LGTM, please revert version change in package.json

PR #891, opened 2022-10-27
Fix P1 issues

ankhyk:


lib/generator/exampleGenerator.ts
```diff
@@ -174,6 +174,12 @@ export default class Generator {
       }
     } else {
       example = {
+        title: specItem.content.summary
```
raych1: Firstly use the correspondent element key within `x-ms-examples`, if it fails to get then use your current logic.

lib/generator/exampleGenerator.ts
```diff
@@ -173,12 +173,18 @@ export default class Generator {
...
+      const title = xMsExampleKeys.length > 0 ? xMsExampleKeys[0] : "";
       example = {
-        title: specItem.content.summary
-          ? specItem.content.summary
-          : specItem.content.description
-          ? specItem.content.description
-          : operationId,
+        title:
+          title.length > 0
+            ? title
```
raych1: Title should include the rule name, otherwise it will be duplicated. i.e. {title}_maximumSet and {title}_minimumSet

raych1 (commented):
Can you add changelog?

raych1 (approved):

PR #907, opened 2022-11-18
enable codeQL

ankhyk:


azure-pipelines.yml
```diff
@@ -1,6 +1,9 @@
 pool:
   vmImage: "Ubuntu 20.04"
 
+variables:
```
raych1, line 4: Keep this one is enough and removed others.

raych1 (approved):

Issue #901, opened 2022-11-08
Ignore leading slash for scope parameter


raych1:
See below issue for details.
https://github.com/Azure/azure-sdk-tools/issues/4420

Issue #893, opened 2022-11-02
INVALID_TYPE isn't clear to user in a specific case


raych1:
Detail issue is tracked [#4519](https://github.com/Azure/azure-sdk-tools/issues/4519)

Issue #882, opened 2022-10-18
Support path validation


ankhyk:
From the [definition](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#paths-object) of path, the parameter of path should be included by `{}`, OAV should report an error if one of `{}` is missing.

PR #911, opened 2022-11-24
Fix issues 882 893 901

ankhyk:
#882 #893 #901


test/semanticValidatorTests.ts
```diff
@@ -100,6 +100,16 @@ describe("Semantic validation", () => {
...
       );
     });
+    // path brackets
+    it("should fail when path brackets are deployed as a single", async () => {
+      const specPath = `${testPath}/semanticValidation/specification/validateSwaggerSchema/internalErrors/pathBrackets.json`;
+      const result = await validate.validateSpec(specPath, undefined);
+      assert(result.validityStatus === false);
+      assert.strictEqual(
+        result.validateSpec?.errors?.[0].code,
+        constants.ErrorCodes.InternalError.name
```
raych1: Can we turn this error to a new error code instead of internal error as it's swagger error and user has to fix it?

lib/swaggerValidator/modelValidator.ts
```diff
@@ -359,7 +359,8 @@ export class SwaggerExampleValidator {
           // where there will be duplicate forward slashes in the url.
           if (
             pathTemplate.charAt(pathTemplate.indexOf(`${parameter.name}`) - 2) === "/" &&
-            parameterValue.startsWith("/")
+            parameterValue.startsWith("/") &&
+            parameter.name !== "scope"
```
raych1: Will parameter name be other word, such as resourceScope, scopes, etc? Can you scan existed swaggers and check the result?

test/modelValidatorTests.ts
```diff
@@ -96,19 +96,25 @@ describe("Model Validation", () => {
...
-      } catch (err) {
-        assert.strictEqual(err.code, "REQUEST_VALIDATION_ERROR");
-        assert.strictEqual(err.innerErrors[0].code, "DOUBLE_FORWARD_SLASHES_IN_URL");
-      }
+      const result = await validate.validateExamples(specPath, operationIds, {
+        consoleLogLevel: "off",
+      });
+      assert(
+        result.length === 1,
+        `swagger "${specPath}" with operation "${operationIds}" contains passed incorrectly.`
```
raych1: Wording of 'contains passed incorrectly' is confusing, can you change to 'contains validation errors'?

lib/swaggerValidator/semanticValidator.ts
```diff
@@ -546,6 +546,16 @@ export class SwaggerSemanticValidator {
           this.addErrorsFromErrorCode(errors, url, meta, operation);
         }
       },
+      onResponse: (response) => {
+        if (response.schema !== undefined) {
+          if (response.schema.noRefWithTypeObject === true) {
+            const meta = getOavErrorMeta("OBJECT_MISSING_REQUIRED_PROPERTY_DEFINITION", {
+              property: "$ref",
```
raych1: Can you update meta.message to `Missing required property definition: $ref when schema type is 'object'`?

lib/transform/resolveNestedDefinitionTransformer.ts
```diff
@@ -25,6 +25,9 @@ export const resolveNestedDefinitionTransformer: SpecTransformer = {
       if (refSelf !== undefined) {
         schema[refSelfSymbol] = refSelf;
       }
+      if (schema.type === "object" && schema.properties === undefined) {
```
raych1: Does schema here already include the value extracted from $ref?

test/semanticValidatorTests.ts
ankhyk: changed to MISSING_PATH_PARAMETER_DECLARATION

lib/swaggerValidator/modelValidator.ts
ankhyk: I use the x-ms-skip-url-encoding property to skip reporting error

lib/transform/resolveNestedDefinitionTransformer.ts
```diff
@@ -15,6 +15,9 @@ export const resolveNestedDefinitionTransformer: SpecTransformer = {
       if (s === undefined || s === null || typeof s !== "object") {
         return;
       }
+      if (s.type === "object" && s.properties === undefined) {
```
raych1: Do properties here already combine the nested definitions?

lib/swaggerValidator/modelValidator.ts
```diff
@@ -765,10 +770,14 @@ export class SwaggerExampleValidator {
     }
 
     if (invalidTypeError) {
-      const meta = getOavErrorMeta(ErrorCodeConstants.INVALID_TYPE as any, {
+      let meta = getOavErrorMeta(ErrorCodeConstants.INVALID_TYPE as any, {
         type: expectedType,
         data: actualType,
       });
+      if (bodySchema.noRefWithTypeObject === true) {
```
raych1: How about the model within body sub level properties?

lib/transform/pathRegexTransformer.ts
```diff
@@ -111,6 +111,9 @@ export const pathRegexTransformer: SpecTransformer = {
     traverseSwagger(spec, {
       onPath: (path, pathTemplate) => {
         let pathStr = pathTemplate;
+        if ([...pathStr.matchAll(/\{/g)].length !== [...pathStr.matchAll(/\}/g)].length) {
```
raych1, line 114: It would be better to only report this error in semantic validation. Can we move this code to there?

lib/transform/pureObjectTransformer.ts
```diff
@@ -9,6 +9,7 @@ export const pureObjectTransformer: GlobalTransformer = {
         (sch.properties === undefined || Object.keys(sch.properties).length === 0) &&
         sch.additionalProperties === undefined &&
         sch.additionalPropertiesWithObjectType !== true
+        // sch.noRefWithTypeObject !== true
```
raych1: Delete this line.

lib/swagger/swaggerTypes.ts
```diff
@@ -271,6 +271,9 @@ export interface Schema extends BaseSchema {
 
   // Additional property support
   additionalPropertiesWithObjectType?: boolean;
+
+  // Type object exists but reference missing
```
raych1:
```suggestion
  // Type object exists but reference missing, this is only used to validate the case of $ref is missing in response body definition
```

ChangeLog.md
```diff
@@ -2,6 +2,9 @@
 
 ## 01/30/2023 3.2.4
 
+- SemanticValidator&ModelValidator - Clear INVALID_TYPE to user in a specific case
```
raych1: Can you bump version to 3.2.5 then add these changes to new version?

lib/transform/resolveNestedDefinitionTransformer.ts
```diff
@@ -11,10 +11,17 @@ export const resolveNestedDefinitionTransformer: SpecTransformer = {
...
+    const visitNestedDefinitions = (
+      s: Schema | undefined,
+      ref?: string,
+      isResponseSchema?: boolean
+    ) => {
       if (s === undefined || s === null || typeof s !== "object") {
         return;
       }
+      if (isResponseSchema && s.type === "object" && s.properties === undefined) {
+        s.noRefWithTypeObject = true;
```
raych1: Can you rename `noRefWithTypeObject`? This is confuse as it validates schema doesn't have properties defined rather than no $ref.

ChangeLog.md
raych1:
```suggestion
- ModelValidator - Clarify message of INVALID_TYPE error when response body doesn't define schema
```

lib/transform/pathRegexTransformer.ts
ankhyk: - This line is for reporting specific error instead of dependency's error in `path._pathRegex`
- The runtime of `pathStr.matchAll` is so fast that it does not affect the overall runtime

lib/swaggerValidator/modelValidator.ts
ankhyk: AJV will only validate the response body property type, the other model can be skipped.

PR #948, opened 2023-01-18
support scope in static generator

jordan13game:
https://github.com/Azure/azure-sdk-tools/issues/4819


test/apiScenario/__snapshots__/apiScenarioGeneratorTest.ts.snap
```diff
@@ -1262,6 +1262,7 @@ Object {
 
 exports[`ApiScenarioGenerator generate api scenario from swagger - storage 1`] = `
 Object {
+  "authentication": undefined,
```
leniatgh, line 1265: if undefined, we should not output this field

leniatgh (approved):
looks good, just one comment

test/apiScenario/__snapshots__/apiScenarioGeneratorTest.ts.snap
jordan13game: It won't output in yaml, but it should define in object to fix the order in the output.

leniatgh (approved):

PR #960, opened 2023-02-07
Revert node requirement to >=10

ankhyk:
- @apidevtools/swagger-parser has problem supporting node <=14(with npm < 7): [swagger-parser 10.1.0 is broken when used with npm < 7 or using --legacy-peer-deps](https://github.com/ajv-validator/ajv-draft-04)
```
`-- @apidevtools/swagger-parser@10.1.0
  `-- ajv-draft-04@1.0.0
```
- @apidevtools/json-schema-ref-parser has fix the compatibility, so we can revert @apidevtools/swagger-parser to swagger-parser


raych1:
@mikeharder , we have to revert this as it causes runtime issue on node <=14.

package.json
```diff
@@ -40,7 +40,7 @@
     "path-to-regexp": "^6.2.1",
     "postman-collection": "^4.1.4",
     "reflect-metadata": "^0.1.13",
-    "@apidevtools/swagger-parser": "^10.1.0",
+    "swagger-parser": "^10.0.3",
```
mikeharder:
```suggestion
    "@apidevtools/swagger-parser": "10.0.3",
```

`swagger-parser` is just a redirect to `@apidevtools/swagger-parser`, so I think it's better to directly depend on the latter.  The only reason `swagger-parser@^10.0.3` even works, is because the maintainers didn't publish `swagger-parser@10.1.0` when they published `@apidevtools/swagger-parser@10.1.0`.  I'm not sure if this was intentional or a mistake.

If you depend on `swagger-parser@^10.0.3`, and the maintainers do publish a new version of `swagger-parser` in the future, you will be broken again.

package.json
```diff
@@ -132,7 +132,7 @@
     "li": "npm install @ts-common/local-install --no-save && local-install"
   },
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=10.11.0"
```
mikeharder: What versions of Node do you actually want to guarantee are compatible with `oav`?  Personally, I would use Node 14, since it's the oldest version of Node still supported.

Ideally, you should also run automated tests on all the versions of Node you claim to support.  So if `engines` specifies `node >= 14.0.0`, you should run tests on Node 14, 16, and 18.

mikeharder (changes requested):

package.json
ankhyk: Thanks for reminding, I'll lock @apidevtools/swagger-parser to v10.0.3

package.json
raych1: Let's keep node 12. Then later on we can retire 12 support with major version upgrade. we should run tests against different node version in CI pipeline.

ChangeLog.md
```diff
@@ -2,6 +2,7 @@
 
 ## 02/03/2023 3.2.5
 
+- Revert node requirement to >=10
```
mikeharder:
```suggestion
```

mikeharder (approved):
Approved pending my suggestion

package.json
mikeharder: I personally think you can drop support for Node 12 (or any EOL Node version) in either a major or a minor/patch upgrade.

Per strict semver, dropping support for a Node version is probably considered a breaking change.

However, one could also argue:
* Packages should drop support for Node versions when they stop testing on those versions
* It's unreasonable to expect package maintainers to test on Node versions very long after they are EOL
* Packages should not be forced to upgrade their major versions when Node versions go EOL

mikeharder (approved):

raych1:
@keryul , can you add change log saying 'lock apidevtools/swagger-parser@10.0.3 as 10.1.0 brings conflict of ajv package version'

raych1 (approved):

PR #965, opened 2023-02-19
log additionalProperties invalid type error only

raych1:


lib/liveValidation/operationValidator.ts
```diff
@@ -276,17 +299,32 @@ export const schemaValidateIssueToLiveValidationIssue = (
...
-      const isMissingRequiredProperty = issue.code === "OBJECT_MISSING_REQUIRED_PROPERTY";
-      const isBodyIssue = path.startsWith(".body");
-
       if (issue.code === "MISSING_RESOURCE_ID") {
         // ignore this error for sub level resources
         if (path.includes("properties")) {
           skipIssue = true;
           return "";
         }
+      } else if (issue.code === "INVALID_TYPE" && isArmCall !== undefined && isArmCall === false) {
```
jianyexi:
```suggestion
      } else if (issue.code === "INVALID_TYPE" && isArmCall === false) {
```

lib/liveValidation/operationValidator.ts
```diff
@@ -262,7 +282,10 @@ export const schemaValidateIssueToLiveValidationIssue = (
   input: SchemaValidateIssue[],
   operation: Operation,
   ctx: SchemaValidateContext,
-  output: LiveValidationIssue[]
+  output: LiveValidationIssue[],
+  info: OperationContext,
```
jianyexi: seem the 'info' here isn't specific, could you rename it ?

lib/liveValidation/operationValidator.ts
xiaoxuqi-ms: this config is optional,

lib/liveValidation/operationValidator.ts
```diff
@@ -276,17 +299,32 @@ export const schemaValidateIssueToLiveValidationIssue = (
...
-      const isBodyIssue = path.startsWith(".body");
-
       if (issue.code === "MISSING_RESOURCE_ID") {
         // ignore this error for sub level resources
         if (path.includes("properties")) {
           skipIssue = true;
           return "";
         }
+      } else if (issue.code === "INVALID_TYPE" && isArmCall !== undefined && isArmCall === false) {
+        if (issue.schemaPath.includes("additionalProperties")) {
```
xiaoxuqi-ms, line 309: what is the rule for additionalProperties, why we can ignore the invalid_type for additionalProperties?

lib/liveValidation/operationValidator.ts
raych1: Renamed.

lib/liveValidation/operationValidator.ts
raych1: This work is what we discussed in last two sync up meeting. The context is: oav new version added the `additionalProperties` type validation which appears to be a breaking change to service team. There's no easy approach to identify the service who has inconsistence between annotations of `additionalProperties` and API implementation. 

We don't want to introduce LSI by this new version release.  That's why we have this PR. Feel free to ping me offline for further questions.

xiaoxuqi-ms (approved):

PR #971, opened 2023-04-12
Update dependency on `core-http`

scbedd:
So that we can close the gap in `validate-v2` package. 

Within `validate-v2`, there are 3 deps bringing in the high pri vuln of `xml2js`:

- `oav`
- `@azure/ms-rest-azure-js`
- `@azure/ms-rest-js`

This PR patches `oav`. I have bumped the dependency on the other two within the `validate-v2` package.json. Just awaiting the updated dep on oav.


scbedd:
There is an additional CVE for `requests` from `ms-rest`, but that has no released solution yet. Going to reach out to the JS folks to talk to them.

benbp (approved):

PR #974, opened 2023-04-18
Update Regression Snapshots

scbedd:
`oav` version has changed.

Is this a safe update? Is it not? very difficult to tell as the latest regression-full builds _before_ my changes were failing.


scbedd:
This PR needs to be updated after the correct version reversion is completed. Holding off on merge till then.

scbedd:
Merged [this pr](https://github.com/Azure/oav/pull/981) for this update.

PR #985, opened 2023-07-05
Re-Enable AdditionalProperties enforcement

scbedd:
- Version of `oav` being used in `Azure-rest-api-specs` = `3.2.2`
- This [PR](https://github.com/Azure/oav/pull/965/files) introduced as version `3.2.6` makes these `additionalProperties` validations a logged warning by default.
- I have two situations, `oav` when used in `azure-rest-api-specs` SHOULD throw on `additionalProperties` validation errors. `oav` when used in the `validate-v2` RPaaS LiveValidation image, should only **log** these as warnings.
- Prior to this PR, `oav` will not throw on these `additionalProperties` errors at all, as `isArmCall` is DEFAULTED to `false`, which effectively means all of these errors become warnings.
- This PR makes `isArmCall` actually matter, as this `log as warning` code-path will now either skip or not based on the value of `isArmCall` in the LiveValidator. We [have set isArmCall](https://devdiv.visualstudio.com/DevDiv/_git/openapi-platform/pullrequest/482367) in the `validate-v2` `LiveValidator` default setup to allow `RPaaS` folks to continue to have `additionalProperties` errors logged as warnings.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 07/05/2023 3.2.11
+
+- Re-enable `additionalProperties` validation where the `liveValidator` configuration has `isArmCall` set to `false`. The version of `oav` that is being used to validate `azure-rest-api-specs` has this behavior enabled, so this is not a breaking change.
+
 ## 06/15/2023 3.2.10
 
 - Patch `byte` type validation to properly decode and re-encode a base64 string. A `byte` formatted body's validity has **nothing** to do with whether or not one can successfully encode it to ascii.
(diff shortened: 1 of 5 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 07/05/2023 3.2.11
+
+- Re-enable `additionalProperties` validation where the `liveValidator` configuration has `isArmCall` set to `false`. The version of `oav` that is being used to validate `azure-rest-api-specs` has this behavior enabled, so this is not a breaking change.
```
weshaggard, line 5: Perhaps I'm missing something but this comment isArmCall=false doesn't seem to match the condition in the code.

scbedd: That's the point. `isArmCall` defaults to `false`. We really want that condition to only activate in the case of the context of an arm call. By changing it to `isArmCall==true` that specialcase only applies when we are actually validating an armRequest. Otherwise, we don't activate the special case behavior anymore. This means that we can continue to use a standard `oav` call in `azure-rest-api-specs` and CATCH these `additionalproperties` errors that would otherwise have been ignored.

weshaggard (approved):

PR #996, opened 2023-08-25
[Test] Support readOnly on required properties

ankhyk:
just a test PR for https://github.com/Azure/azure-sdk-tools/issues/6840

PR #1004, opened 2023-09-29
Resolve Erroneous `additionalProperty` error on `refWithReadOnly`

semick-dev:
PR #996 fixed a real bug in OAV. However, unbeknownst to me, and hidden in #1003 snapshots too is the fact that by injecting our property `refWithReadOnly`, we're actually beginning to _fail_ semantic validation in certain places. This is occurring because _we've actually added a property to the schema_ by injecting `refWithReadOnly`. Given the schemaValidator doesn't understand it, 

This PR adds to the 'selective ignore' list that is maintained to bypass certain errors that we want to ignore. Here's the thing. I really think that the true solution is to fix this in the schema validator.

I'm betting that there is a way to add the new property to the acceptable schema by default, but I need to find out. In the meantime, this is what it would look like to just ignore this specific error, let's just check if there is an unblock available for us in case the schema validator route proves arduous.


test/individuaSemanticValidationTest.ts
```diff
@@ -0,0 +1,22 @@
...
+import assert from "assert";
+import * as validate from "../lib/validate";
+
+const testPath = __dirname.replace("\\", "/");
+
+describe("Semantic validation", () => {
+  it("a valid minimal swagger should pass semantic validation", async () => {
+
+    
+    const specPath = `${testPath}/modelValidation/swaggers/specification/refWithReadOnlyProperyError/cognitiveservices.json`;
```
semick-dev: Move to `semanticValidatorTests`

scbedd:
Closing this PR, as the prettier fixes are more significant than I wanted to keep around.

scbedd:
Re-opening. This is actually correct! Applying lint fixes locally came up with same result.

scbedd (approved):

PR #1005, opened 2023-10-05
Add Regression Pipeline

semick-dev:
This PR adds a pipeline that is intended to eliminate the manual effort associated with a regression comparison of `oav` runs after code changes.


scbedd:
Closing.

PR #1019, opened 2023-11-15
Resolve Regressions on on behavior of `anyOf` due to updates on `refWithReadOnly`

scbedd:
With the addition #996 and #1004 , we introduced regressions on a few specs across the azure-rest-api-specs repository.

They were:
- specification/automation/resource-manager/Microsoft.Automation/preview_2017-05-15/preview/job.json
- specification/automation/resource-manager/Microsoft.Automation/stable/2019-06-01/job.json
- specification/automation/resource-manager/Microsoft.Automation/stable/2022-08-08/job.json
- specification/datamigration/resource-manager/Microsoft.DataMigration/stable/2018-04-19/definitions/MigrationValidation.json
- specification/maps/data-plane/Search/preview_1.0/search.json
- specification/scvmm/resource-manager/Microsoft.ScVmm/preview/2020-06-05-preview/scvmm.json
- specification/scvmm/resource-manager/Microsoft.ScVmm/preview/2022-05-21-preview/scvmm.json
- specification/scvmm/resource-manager/Microsoft.ScVmm/preview/2023-04-01-preview/scvmm.json
- specification/scvmm/resource-manager/Microsoft.ScVmm/stable/2023-10-07/scvmm.json

We started throwing new errors of various types due to our code changes. The specs are just fine, so these were regressions that needed to be fixed.

After a bunch of digging, I finally have the understanding to understand what was going on.

in #996 , we added a property to the `Schema` interface, which allows us to _know_ when we resolve a `ref` node that happens to have `readonly` on it. It does this by setting a new property `refWithReadOnly` to `true` as each `ref` node in a schema is resolved. (This property addition happens at runtime while we ensure all the `ref` nodes resolve)

However, by adding this property to every schema before the schema is passed to swagger validation, we accidentally introducing weird swagger validation errors. This is because we are actually not passing a schema that RECOGNIZES this `refWithReadOnly` value that we injected.  We pull the `swagger` schema from `import swaggerSchemaDoc from "@autorest/schemas/swagger-extensions.json"`. This is the json schema that is used to construct a schema validation function for a generic "swagger, and includes the swagger extensions that microsoft users in @autorest.

The "source of truth" schema used to validate a swagger spec does NOT recognize the presence of the property that we added in #996. That is the source of all these wonky new errors in the files mentioned above.

While digging for understanding to see if I should upstream any properties to the swagger extension json....I realized that `x-ms` prefixed values are pretty much ignored by the swagger validation. I thought that I'd rename `refWithReadOnly`  to one of these `x-ms` prefixed properties, all our wonky schema issues go away.

This PR merely shifts the `refWithReadOnly` identifier **to be a property prefixed by x-ms**. Specifically `"x-ms-readonly-ref"`

Regression run for this PR: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3262880&view=results


scbedd:
I am extremely pleased to say that while the regression tests "failed" due to the presence of a diff....it's a great diff!

This was the error on the previous stable version of `oav` `3.2.11`:

image

This was from example file: `specification/cognitiveservices/data-plane/ContentSafety/stable/2023-10-01/examples/AddOrUpdateBlocklistItems.json`.

This error no longer occurs on `3.3.1`. It doesn't occur because of the changes to `readonly` that we made. One cannot provide a property that defined to be `readonly`. That property is something that will be _returned_ to you populated by the API. In the example that is no longer erroring, the example provides a list of items for operationId `TextBlocklists_AddOrUpdateBlocklistItems`. The user wouldn't provide the `TextBlockListItemId`, they'll provide the _other_ metadata  and be _returned_ the `TextBlockListItemId`. This is expressed in swagger by setting `TextBlockListItemId` to readonly.

Link to example file [no longer erroring.](https://github.com/Azure/azure-rest-api-specs/blob/0dd49a444195fef7f3555cad038cb7665cbd928c/specification/cognitiveservices/data-plane/ContentSafety/stable/2023-10-01/examples/AddOrUpdateBlocklistItems.json#L10)

I'm considering this ready to go in! NO OTHER REGRESSIONS PRESENT!

lib/swagger/swaggerTypes.ts
```diff
@@ -18,6 +18,7 @@ import {
   xmsPaths,
   xmsSkipUrlEncoding,
   xNullable,
+  xMsReadonlyRef,
```
weshaggard: nit: but all other xms variables use a lower case M.

scbedd: **sigh**

you're absolutely right :D 

Will update.

weshaggard (approved):

PR #1024, opened 2024-01-24
Remove mockjs from dependencies

### LINKED ISSUE #1023: Replace dependency on `mock.js`
Need to find an alternative, or generate a version without the dependency.

scbedd:
Resolves #1023 

While all the tests still pass, we have a GAP with this solution. Specifically, what `mockjs` was doing for us was generating a string that _matched_ [a spec parameter that had a custom pattern](https://github.com/Azure/azure-rest-api-specs/blob/c75671ae0a5d82d7d96fda75150478e8581408d8/specification/web/resource-manager/Microsoft.Web/stable/2016-08-01/WebApps.json#L18284)

So, with this removal, we stop being able to properly generate an example for a parameter that maintains a custom pattern.

Per discussion with Wes, we are falling back to substituting a string which includes the necessary regex in the generated example.

image

We need to 

- [x] Add a test which exercises this custom pattern
- [x] Resolve the dependency on `mockjs` for this generation.


test/exampleGeneratorTests.ts
```diff
@@ -80,6 +80,7 @@ describe("test generate example", () => {
     ["sql", "package-pure-2020-02-preview"],
     ["signalr", "package-2020-05-01"],
     ["eventgrid", "package-2020-06"],
+    ["agrifood", "package-preview-2023-06"]
```
konrad-jamrozik, line 83: Just curious: why we are adding `agrifood` samples in this PR?

konrad-jamrozik (approved):

test/exampleGeneratorTests.ts
scbedd: It includes a `pattern` usage in a parameter for the operation present within. Originally, when I removed the `mockerjs` dependency, these tests did **not** fail because none of the specs being exercised included a parameter that utilized the `pattern` property.

Now it does, so if the generation actually **failed** we'd know.

PR #1025, opened 2024-02-12
Port `arm`-specific changes into `develop`

scbedd:
After going back and forth with the ARM folks a bit, I've realized that a previous update to `oav` was incorrect. Not only that, but we have some additional suppression logic that applies only to the ARM team ported in this PR as well.

This PR:

- Enables `arm-id` format validation in all cases.
  - `INVALID_FORMAT` errors will be suppressed and only logged when `isArmCall == false`. In practice, `isArmCall = false` during `liveValidation`, which only applies to the ARM validation image currently.
- Resolve erroneously suppressed `additionalProperties` errors in `operationValidator.ts`. `isArmCall=true` in all cases of `Model Validation` and `Semantic Validation`, which means this suppression was backwards before this PR. A full regression build of this changeset will be checked against `azure-rest-api-specs` to see if anything has slipped through while this is disabled.
- Add unit test to check suppression of `INVALID_FORMAT` error during live validation.

### DIFF
--- a/lib/liveValidation/operationValidator.ts
+++ b/lib/liveValidation/operationValidator.ts
@@ -332,7 +332,7 @@ export const schemaValidateIssueToLiveValidationIssue = (
       } else if (
         issue.code === "INVALID_FORMAT" &&
         isArmCall === false &&
-        issue.message.indexOf("Object didn't pass validation for format arm-id") > -1
+        issue.message.includes("Object didn't pass validation for format arm-id")
       ) {
         skipIssue = true;
         if (logging) {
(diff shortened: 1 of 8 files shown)


lib/liveValidation/operationValidator.ts
```diff
@@ -327,6 +329,22 @@ export const schemaValidateIssueToLiveValidationIssue = (
           }
           return "";
         }
+      } else if (
+        issue.code === "INVALID_FORMAT" &&
+        isArmCall === false &&
+        issue.message.indexOf("Object didn't pass validation for format arm-id") > -1
```
konrad-jamrozik: nit: you could replace this with 
`issue.message.includes("Object didn't pass validation for format arm-id")`

konrad-jamrozik (approved):

lib/liveValidation/operationValidator.ts
scbedd:
```suggestion
        issue.message.includes("Object didn't pass validation for format arm-id")
```

PR #1028, opened 2024-03-27
Re-enable arm-id validation

wwendyc:
Re-enables the arm-id validation in all cases, even when `isArmCall` is `false`. The special handling for this error will be done in RPaaS's codebase.


scbedd (approved):
LGTM. Thanks Wendy!

PR #1029, opened 2024-05-07
bump node engine version

wwendyc:

PR #1032, opened 2024-06-01
Support all current LTS Node versions (>=18.0.0)

mikeharder:
- Overrides #1029

We should support all current LTS Node versions (>=18.0.0) for maximum compatibility.  Matches other packages like `@typespec/compiler`:

https://github.com/microsoft/typespec/blob/98f8d62e3fa8a01f9a594853b27985d071a03355/packages/compiler/package.json#L50

Detected failure when trying to add `oav` to `azure-rest-api-specs/package.json`:

https://github.com/Azure/azure-rest-api-specs/actions/runs/9327612995/job/25677733715#step:4:13

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 05/31/2024 3.3.6
+
+- Support all current LTS Node versions (>=18.0.0)
+
 ## 05/7/2024 3.3.5
 
 - Bump Node engine version requirement for security vulnerability.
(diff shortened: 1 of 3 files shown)


scbedd (approved):
I'll approve this, merge it. And follow up with @wwendyc to be certain we're all good.

scbedd:
Just FYI @mikeharder I've got a bigger refactor [over here](https://github.com/Azure/oav/pull/1031) which gets oav compatible security wise. I'll ship a version of oav that includes this change later today.

PR #1036, opened 2024-06-12
Bump oav version, deps, and changelog

scbedd:


mikeharder (approved):

scbedd:
Bypassing merge requirements. I want a successful release before I delete the previous existing builds and releases. They still exist and are blocking merge because of me wanting to keep them arond just in case.

PR #1042, opened 2024-07-12
Don't generate examples for `password` with a random string

### LINKED ISSUE #1041: Update oav generated sample values for password to use a secret placeholder instead of a random string that looks like it could be a password
https://github.com/Azure/azure-rest-api-specs/pull/29673#discussion_r1665305465 service team is using oav to generate the rest api examples, but codegen will use those samples to generate the SDK sample, which causes sdk credential scan failure https://github.com/Azure/azure-sdk-for-js/issues/30232

scbedd:
Resolves #1041 

Instead we are generating with the value `<a-password-goes-here>`.

This makes it obvious to users of the example, without breaking any workflows. I'll take care of adding this value to credscan necessary.

"SecretPlaceHolder" seemed almost arbitrary to me, and I figured I'd encode a bit more data in the value. If we just want to use that original string then I'll just do that.

### DIFF
--- a/lib/generator/mocker.ts
+++ b/lib/generator/mocker.ts
@@ -34,7 +34,7 @@ export default class Mocker {
     }
 
     if (paramName === "password" || paramName === "pwd" || paramName === "adminPassword") {
-      return "SecretPlaceHolder";
+      return "<a-password-goes-here>";
     }
 
     if (paramSpec.format === "date") {
--- a/lib/generator/mocker.ts
+++ b/lib/generator/mocker.ts
@@ -29,11 +29,13 @@ export default class Mocker {
   }
 
   private generateString(paramSpec: any, paramName: string) {
+    const pwdParams = ["password", "pwd", "adminPassword"];
+
     if (paramSpec.name === "subscriptionId") {
       return uuid.v4().toUpperCase();
     }
 
-    if (paramName === "password" || paramName === "pwd" || paramName === "adminPassword") {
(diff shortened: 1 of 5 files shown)


lib/generator/mocker.ts
```diff
@@ -33,6 +33,10 @@ export default class Mocker {
       return uuid.v4().toUpperCase();
     }
 
+    if (paramName === "password" || paramName === "pwd" || paramName === "adminPassword") {
```
scbedd: Scanned the repo, looking for other properties that might hit the same issue.

kurtzeborn (approved):

lib/generator/mocker.ts
```diff
@@ -29,10 +29,16 @@ export default class Mocker {
...
 
   private generateString(paramSpec: any, paramName: string) {
+    const pwdParams = ["password", "pwd", "adminPassword"];
+
     if (paramSpec.name === "subscriptionId") {
       return uuid.v4().toUpperCase();
     }
 
+    if (pwdParams.includes(paramName)) {
+      return "<a-password-goes-here>";
```
kurtzeborn, line 39: This is really great as it will eliminate a whole class of bogus bug bounty requests for things that *look like* leaked passwords but indeed are not valid passwords.  This will make it obvious to everyone that these are not leaked.  Great improvement!

kurtzeborn (approved):

PR #1044, opened 2024-08-30
Add Passed opeartions to the report

ms-zhenhua:
Some users complained that they do not know which operations are tested successfully from the report. This PR adds the passed operations to the report.

image

### DIFF
--- a/lib/report/generateReport.ts
+++ b/lib/report/generateReport.ts
@@ -3,6 +3,7 @@ import * as path from "path";
 import * as Mustache from "mustache";
 import {
   OperationCoverageInfo,
+  OperationMeta,
   TrafficValidationIssue,
   TrafficValidationOptions,
 } from "../swaggerValidator/trafficValidator";
@@ -52,9 +53,18 @@ export interface ErrorDefinition {
   link?: string;
 }
 
+export interface ValidationPassOperationsFormatInner extends OperationMeta {
+  readonly key: string;
+}
+
+export interface ValidationPassOperationsFormat {
+  readonly operationIdList: ValidationPassOperationsFormatInner[];
+}
+
 export interface OperationCoverageInfoForRendering extends OperationCoverageInfo {
   specLinkLabel?: string;
   validationPassOperations?: number;
+  validationPassOperationList: ValidationPassOperationsFormat[];
   generalErrorsInnerList: TrafficValidationIssueForRenderingInner[];
 }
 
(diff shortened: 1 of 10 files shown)


lib/report/generateReport.ts
```diff
@@ -52,9 +53,18 @@ export interface ErrorDefinition {
   link?: string;
 }
 
+export interface ValidationPassOperationsFormatInner extends OperationMeta {
```
scbedd, line 56: `ValidationPassOperationsFormatInner` is not a descriptive name. I believe that you're creating a new data type to easily walk the different "types" of the operation via the `key` value. I believe you're repivoting by whatever the value of this `key` is, then you've got a list of operation ids under _that_. Maybe `CategorizedOperationMeta` or something like that?

Can you please offer a comment describing what an `OperationMeta` will look like in context with a code comment? (EG pseudo or real json / operation id) so a reader can easily infer WHY the split on `_` works.

lib/report/generateReport.ts
```diff
@@ -201,18 +211,77 @@ export class CoverageView {
           runtimeExceptions: element.runtimeExceptions,
         });
       });
+
+      const generalErrorsInnerOrigin = this.validationResultsForRendering.filter((x) => {
```
scbedd, line 215: Nit: `generalErrorsInnerOrigin` means "any validation result that has errors." Why not call this `failingValidationResults`? 

Effectively:

- `validationResultsForRendering` (also a long name but refactors aren't free so if it works as-is leave it IMO)
- `errorValidationResults`
- `passValidationResults`
- Where error and pass added together equals `validationResultsForRendering`.

lib/report/generateReport.ts
```diff
@@ -201,18 +211,77 @@ export class CoverageView {
...
+        let passOperations: ValidationPassOperationsFormatInner[] = element.coveredOperationsList
+          .filter((item) => errorOperationIds.indexOf(item.operationId) === -1)
+          .map((item) => {
+            return {
+              key: item.operationId.split("_")[0],
+              operationId: item.operationId,
+            };
+          });
+
+        const passOperationsInnerList: ValidationPassOperationsFormatInner[][] = Object.values(
```
scbedd, line 238: I don't see anything that jumps out at me in the code block. Once I get a bit more clarity I'll have a better understanding though.

scbedd (changes requested):
1. Have some feedback seeking clarity on the generateReport
2. Please update `package.json`, `package.lock` and `changelog.md` to account for a new MINOR version (this is a new feature). Please target version `3.5.0`. `npm install` to update the `package.lock` after updating the `package.json`.

lib/report/generateReport.ts
ms-zhenhua: the `OperationMeta` is referenced from the [existing code](https://github.com/ms-zhenhua/oav/blob/ed2352f4e9fed02091252140c248f10ca22b9721/lib/swaggerValidator/trafficValidator.ts#L64). `ValidationPassOperationsFormatInner` is similar with [unCoveredOperationsFormatInner](https://github.com/ms-zhenhua/oav/blob/ed2352f4e9fed02091252140c248f10ca22b9721/lib/swaggerValidator/trafficValidator.ts#L68C18-L68C48). The name is defined as `ValidationPassOperationsFormatInner` because there already exists property name [validationPassOperations](https://github.com/Azure/oav/blob/d24383eb554abd7fa656f566d630f8702d49babf/lib/report/generateReport.ts#L57) so that I use the same prefix.

lib/report/generateReport.ts
ms-zhenhua: 1. `generalErrorsInnerOrigin` is not a new property. It already exists in [existing code](https://github.com/Azure/oav/blob/d24383eb554abd7fa656f566d630f8702d49babf/lib/report/generateReport.ts#L238). I moved its position ahead because I need to use the error operations to calculate the passed operations. The logic is `PassedOperations = CoveredOperations - FailedOperations`.
2. I choose to put `passValidationResults` in `coverageResultsForRendering` but not `validationResultsForRendering` for 2 reasons:
           a. there already exists [validationPassOperations](https://github.com/Azure/oav/blob/d24383eb554abd7fa656f566d630f8702d49babf/lib/report/generateReport.ts#L215C11-L215C35)in `coverageResultsForRendering`. This time I just added the operations names in `validationPassOperationList` so that users know which operations have been test.
           b. Since `PassedOperations = CoveredOperations - FailedOperations`, I need to use some Covered Operations info in `coverageResultsForRendering`.

lib/report/generateReport.ts
ms-zhenhua: This code is similar with this[ existing code](https://github.com/Azure/oav/blob/d24383eb554abd7fa656f566d630f8702d49babf/lib/swaggerValidator/trafficValidator.ts#L319), which is used to group passed operations by APIs.

ms-zhenhua:
Hi @scbedd, thank you for your review. I have updated the `package.json`, `package.lock` and `changelog.md`. I also added the explanation for your comments. Could you please have another review?

lib/report/generateReport.ts
scbedd: Frankly I don't think we should keep awkward naming conventions just because it's awkward elsewhere.

The lack of basic comments for complex types describing the WHY of these magic splits is my single biggest frustration of this codebase. "An operationId looks like this, which is why the arbitrary split works" MAKES THE CODE SO MUCH EASIER to walk.

scbedd (approved):
My feedback was offered. Approved to unblock the feature.

PR #1054, opened 2024-10-15
[generate-examples] Include Min/Max in default example titles

### LINKED ISSUE #935: oav generate-examples generated min and max sets with same title
The generated examples have the same title value for both the min and max set for every operation. So, the `cadl compile` links (with the `x-ms-examples`) just one of them in the generated swagger as the other one gets override.

Both the titles in the min and the max set example files should have different titles (maybe with a suffix like _min or _max) so that the CADL links them properly to the swagger during compilation.

mikeharder:
- Fixes #935 
- Uses unique titles when first generating examples
- Still has unrelated bugs where title can be changed when updating existing examples

### DIFF
--- a/lib/generator/exampleGenerator.ts
+++ b/lib/generator/exampleGenerator.ts
@@ -170,7 +170,7 @@ export default class Generator {
     let example;
     console.log(`start generated example for ${operationId}, rule:${rule.ruleName}`);
     if (!this.shouldMock) {
-      example = this.getExampleFromPayload(operationId, specItem);
+      example = this.getExampleFromPayload(operationId, specItem, rule);
       if (!example) {
         return [];
       }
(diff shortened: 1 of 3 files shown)


lib/generator/exampleGenerator.ts
```diff
@@ -186,7 +186,7 @@ export default class Generator {
             ? specItem.content.summary
             : specItem.content.description
             ? specItem.content.description
-            : operationId,
+            : `${operationId}_${rule.exampleNamePostfix}`,
```
mikeharder, line 189: Similar to pattern used for example filenames

scbedd (approved):
I'm glad it was a minor change.

dargilco (approved):

PR #1075, opened 2025-06-24
Fix the issue with being unable to handle identifiers directly trailing a parameter

### LINKED ISSUE #1073: `{scale}x` within path is breaking `oav`
This is the path:

```
/data/collections/{collectionId}/items/{itemId}/crop/{width}x{height}.{format}
```

The issue is that `oav` replaces these parameters with integer keys representing each section of the path, so the above path becomes:

```
'/data/collections/:0/items/:1/crop/:2x:3.:4'
```

And because a `parameter` is `[A-Za-z0-9]`, we are recognizing `2x` as a single parameter identifier.

`path-to-regexp` can handle this by quoting the parameters specifically. So the above string becomes:

```
'/data/collections/:"0"/items/:"1"/crop/:"2"x:"3".:"4"'
```

And `path-to-regexp` will create a regex out of that properly. I just need to deal with the downstream impacts of this new formatting + matching.

scbedd:
Resolves #1073

### DIFF
--- a/lib/transform/pathRegexTransformer.ts
+++ b/lib/transform/pathRegexTransformer.ts
@@ -77,7 +77,7 @@ const buildPathRegex = (
       hasMultiPathParam = true;
     } else {
       hostTemplate = hostTemplate.replace("{" + v + "}", ":" + i);
-      path = path.replace("{" + v + "}", ":" + i);
+      path = path.replace("{" + v + "}", `:${i}(\\d+)`);
     }
   });
 
@@ -88,7 +88,13 @@ const buildPathRegex = (
   // in security patched versions of path-to-regexp (read > 6.2.1), colons must refer to valid parameter names
   // so now we have to escape literal colons that are in the path
   const escapedPath = escapeLiteralColons(processedPath);
-  const regexp = pathToRegexp(escapedPath, keys, { sensitive: false });
+  let regexp: RegExp;
+  try {
+    regexp = pathToRegexp(escapedPath, keys, { sensitive: false });
+  }
+  catch(e) {
+    throw e;
+  }
 
   // restore parameter name
   const _keys: string[] = [];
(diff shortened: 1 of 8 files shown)


lib/transform/pathRegexTransformer.ts
```diff
@@ -88,7 +88,13 @@ const buildPathRegex = (
   // in security patched versions of path-to-regexp (read > 6.2.1), colons must refer to valid parameter names
   // so now we have to escape literal colons that are in the path
   const escapedPath = escapeLiteralColons(processedPath);
-  const regexp = pathToRegexp(escapedPath, keys, { sensitive: false });
+  let regexp: RegExp;
```
scbedd, line 132: will revert to allow crash.

lib/transform/pathRegexTransformer.ts
```diff
@@ -77,7 +85,7 @@ const buildPathRegex = (
       hasMultiPathParam = true;
     } else {
       hostTemplate = hostTemplate.replace("{" + v + "}", ":" + i);
-      path = path.replace("{" + v + "}", ":" + i);
+      path = path.replace("{" + v + "}", `:${i}(\\d+)`);
```
timotheeguerin: Is there any unit test for this that could add a regression check for this

timotheeguerin (approved):

lib/transform/pathRegexTransformer.ts
scbedd: A full regression run is literally a re-run of pretty much the whole `azure-rest-api-specs` repo, so at the very least none of the existing public specs don't have this issue.

I'll add a test for it but I wanted to get the PR out and regression builds running.

scbedd:
We pinned `path-to-regexp@6.2.2`, opted into the security issue. This method is not the approach.

PR #1079, opened 2025-07-17
Support identifier chars after params in path-to-regexp@6.3.0

### LINKED ISSUE #1073: `{scale}x` within path is breaking `oav`

mikeharder:
- fixes #1073

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 07/17/2025 3.6.3
+
+- Follow up on 3.6.1 dealing with edge-cases broken by security fixes in `path-to-regexp@6.3.0`. `oav` now properly handles identifier characters immediately following a parameter - EG `{param1}x{param2}`.
+
 ## 6/19/2025 3.6.2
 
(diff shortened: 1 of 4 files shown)


mikeharder:
@scbedd: Good enough, or should we add tests?

scbedd (approved):
Good enough. Let's get this out. When we move `oav` or actually invest in it we'll begin adding true unit tests.

scbedd:
~~Ah but we need to bump the package and changelog.~~

PR #1083, opened 2025-08-01
MULTIPLE_BODY_PARAMETERS should allow additional content-types

### LINKED ISSUE #1057: MULTIPLE_BODY_PARAMETERS error for application/x-www-form-urlencoded
If an OpenAPI 2.0 document contains an operation which contains multiple parameters in formData and consumes `application/x-www-form-urlencoded`, validation fails with `MULTIPLE_BODY_PARAMETERS`.  If consumes is changed to `multipart/form-data`, validation succeeds without error.  For example:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "Title",
    "version": "1.0"
  },
  "paths": {
    "/": {
      "post": {
        "operationId": "test",
        "consumes": [
          "application/x-www-form-urlencoded"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "first_name",
            "in": "formData",
            "type": "string"
          },
          {
            "name": "last_name",
            "in": "formData",
            "type": "string"
          }
        ],
        "responses": {
          "204": {
            "description": "Example response"
          }
        }
      }
    }
  }
}
```

The OpenAPI document validates and Autorest generates correct code, but oav validation fails.

This issue is similar to #805.

I believe it could be fixed by checking for `application/x-www-form-urlencoded` in addition to `multipart/form-data` at
 
https://github.com/Azure/oav/blob/bf50f770da65ab3c1026a3a6744113c061cb97f4/lib/swaggerValidator/semanticValidator.ts#L527

Thanks for considering,
Kevin

### LINKED ISSUE #1081: MULTIPLE_BODY_PARAMETERS reported on valid swagger
Getting a MULTIPLE_BODY_PARAMETERS error on the following generated swagger:

```json
"consumes": [
          "multipart/mixed"
        ],
        "parameters": [
          ...
          {
            "name": "name",
            "in": "formData",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "in": "formData",
            "required": true,
            "type": "file"
          }
        ],
```

Could it be that the check isnt recognizing the multipart/mixed content type?

Failure on real-world spec: https://github.com/Azure/azure-rest-api-specs/pull/32527/checks?check_run_id=47158084254

mikeharder:
- Fixes #1057 
- Fixes #1081

### DIFF
--- a/test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
+++ b/test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
@@ -12,7 +12,7 @@
           "multipart/mixed"
         ],
         "produces": [
-          "application/json"
+          "multipart/mixed"
         ],
         "parameters": [
           {
--- a/test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
+++ b/test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
@@ -1,0 +1,39 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Title",
+    "version": "1.0"
+  },
+  "paths": {
+    "/": {
+      "post": {
+        "operationId": "test",
+        "consumes": [
+          "multipart/mixed"
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "parameters": [
+          {
+            "name": "name",
+            "in": "formData",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "formData",
+            "required": true,
+            "type": "file"
+          }
+        ],
+        "responses": {
+          "202": {
+            "description": "Example response"
+          }
+        }
+      }
+    }
+  }
+}
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Change Log - oav
 
+## 08/01/2025 3.6.4
+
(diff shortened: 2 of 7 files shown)


test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "paths": {
+    "/": {
+      "post": {
+        "operationId": "test",
+        "consumes": [
+          "multipart/mixed"
+        ],
+        "produces": [
+          "application/json"
```
catalinaperalta: Would it matter if the produces is also multipart/mixed?

catalinaperalta (approved):
LGTM. Thanks @mikeharder!! Just had one question, but I dont think you need to change the example

test/semanticValidation/specification/validateOperation/MULTIPLE_BODY_PARAMETERS/multipartMixed.json
mikeharder: I don't think so, but I'm happy to change the test to match.

scbedd (approved):
This seems reasonable. I'll note that the multipart/mixed is probably a bit special casey for storage, as I don't see refs to it in the swagger parameter docs.

mikeharder:
> This seems reasonable. I'll note that the multipart/mixed is probably a bit special casey for storage, as I don't see refs to it in the swagger parameter docs.

storage-blob is currently the only spec using `multipart/mixed`, but according to @johanste this is the best way to model the actual service in swagger.  If anything downstream of the spec (like SDK generators) fails on `multipart/mixed`, we should detect it then.  We're just shifting this check to the right.

PR #1085, opened 2025-09-09
Convert deployment to releaseJob

weshaggard:
In order for this to be a release job we cannot checkout the source and given we don't need it for anything but the tag which we always pick latest for I've removed that as well.


mikeharder:
LGTM, but I will let Scott approve since he likely knows more about this.

scbedd (approved):

weshaggard:
@mikeharder @scbedd I'll leave this to one of you guys to merge at a point when we are going to do a release so we can validate it actively.

PR #1136, opened 2025-09-24
[dependabot.yml] ignore deps `ajv`, `inversify`, `path-to-regexp`, `*eslint*`, `@types/node`, `*jest*`

mikeharder:


.github/dependabot.yml
```diff
@@ -10,7 +10,9 @@ updates:
       # pinned to the most popular versions
       - dependency-name: "@azure/openapi-markdown"
       - dependency-name: "json-schema-traverse"
-      - dependency-name: "reflect-metadata"
```
mikeharder, line 13: removed to let dependabot upgrade to latest, which might be more compatible with other upgrades

scbedd (approved):

PR #1148, opened 2025-09-24
Remove unused devDependency `npm-run-all`

mikeharder:


mikeharder:
Dependency is actually used, though it's not obvious.  The dependency provides 3 CLI commands (`npm-run-all`, `run-s`, `run-p`), and we use only the last `run-p`.

PR #1190, opened 2025-12-12
[dependabot.yml] Enable for github actions

mikeharder:
- Repo does not use actions yet, but let's enable now so we don't forget
- Reduce frequency of npm updates from daily to weekly

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,16 @@
 version: 2
 updates:
+  # repo does not use GitHub Actions yet, but let's enable dependabot so we don't forget
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,9 +1,16 @@
 version: 2
 updates:
+  # repo does not use GitHub Actions yet, but let's enable dependabot so we don't forget
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
```
mikeharder:
```suggestion
```

raych1 (approved):

